PR #65, opened 2015-01-25
src: log errors from platform

lkundrak:
nl_syserr2nlerr() reduces a lot of platform errors to NLE_FAILURE --
"Unspecific failure" which makes it somehow hard to track down the real reason
behind a failure.

Logging them with level of 4 makes it a little less painful.


thom311:
merged as https://github.com/thom311/libnl/commit/f545295104454090704d29f7bcb27f49bc770f3a

PR #68, opened 2015-02-17
add socket nl_connect_fd() & nl_create_fd()

sagilow:
add socket nl_connect_fd() & nl_create_fd()

-    Added option to create socket (fd) without bind.
	 It is now possible to forward the socket fd to another child process...
	 ...later use nl_connect_fd() to connect to socket from the child process.

-    Added option to disable CLOEXEC even if defined (in socket.h)
	 'nl_socket_enable_cloexec' & 'nl_socket_disable_cloexec'
	 No change to current default behavior.

 Signed-off-by: Sagi Lowenhardt <sagil@infinidat.com>

### DIFF
--- a/include/netlink-private/types.h
+++ b/include/netlink-private/types.h
@@ -78,6 +78,7 @@ struct nl_sock
 	int			s_flags;
 	struct nl_cb *		s_cb;
 	size_t			s_bufsize;
+	int			s_cloexec;
 };
 
 struct nl_cache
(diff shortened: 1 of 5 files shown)


thom311:
Hi Sagi,

I applied your patch, but in a second commit I completely rewrote it :)

Originally I thought that nl_connect() should call nl_connect_fd(). But I no longer think so because nl_connect_fd() should not call bind() or nl_socket_set_buffer_size().
nl_connect_fd() should just allow the user to setup the fd without configuring it further.

Also, at that point I renamed nl_connect_fd() to nl_socket_set_fd(), because that seems like a better name.

for now, I merged it and close this pull request. Let's see how this goes and if necessary we can improve it later. How do you like it?


Thank you,
Thomas

PR #73, opened 2015-03-17
Remove unnecessary include of private linux/if.h

davidchappelle:
Including linux/if.h in netlink/route/link.h causes issues
in cases where libnl is used in conjuntion with other third
party libraries that include net/if.h. Seems to be a long
checkered history of symbol collisions between these two
files. As it turns out, including linux/if.h from within
netlink/route/link.h is actually unecessary. I resurrected
a forgotten path from this thread:

    http://lists.infradead.org/pipermail/libnl/2012-April/000525.html

By removing the include as the patch suggests we can get
around the nuissance of the symbol collisions.

### DIFF
--- a/include/netlink/route/link.h
+++ b/include/netlink/route/link.h
@@ -15,7 +15,6 @@
 #include <netlink/netlink.h>
 #include <netlink/cache.h>
 #include <netlink/addr.h>
-#include <linux/if.h>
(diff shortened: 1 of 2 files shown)


thom311:
Thank you, merged as 50a76998ac36ace3716d3c979b352fac73cfc80a

PR #75, opened 2015-04-17
Revert "route: remove unnecessary include of private linux/if.h"

gustavoz:
It's necessary for IFF_* defines as would be used via
rtnl_link_get_flags() checks against for example IFF_UP.

It breaks bmon for example since it never includes linux/if.h, see:
http://autobuild.buildroot.net/results/296/296bcaf4fd1fadcab71881d6d00a8763bfa91a41/build-end.log

This reverts commit 50a76998ac36ace3716d3c979b352fac73cfc80a.

### DIFF
--- a/include/netlink/route/link.h
+++ b/include/netlink/route/link.h
@@ -15,7 +15,9 @@
 #include <netlink/netlink.h>
 #include <netlink/cache.h>
 #include <netlink/addr.h>
+#ifndef _LINUX_IF_H
 #include <linux/if.h>
+#endif
 #include <sys/types.h>
 
 #ifdef __cplusplus
--- a/include/netlink/route/link.h
+++ b/include/netlink/route/link.h
@@ -15,6 +15,7 @@
 #include <netlink/netlink.h>
 #include <netlink/cache.h>
 #include <netlink/addr.h>
+#include <linux/if.h>
 #include <sys/types.h>
 
 #ifdef __cplusplus
--- a/src/nl-link-set.c
+++ b/src/nl-link-set.c
@@ -9,7 +9,6 @@
  * Copyright (c) 2003-2010 Thomas Graf <tgraf@suug.ch>
  */
 
-#include <linux/if.h>
 #include <netlink/cli/utils.h>
 #include <netlink/cli/link.h>
 


thom311:
+#ifndef _LINUX_IF_H
 #include <linux/if.h>
+#endif

_LINUX_IF_H is the include guard of "linux/if.h", so this check has basically no effect.


Maybe you wanted:
+#ifndef _NET_IF_H
 #include <linux/if.h>
+#endif

but that doesn't solve the issue from 50a76998 in a nice way, because:
- it relies on "net/if.h" to have include guard _NET_IF_H
- it still only works if you include "net/if.h" before "route/link.h".


I think there is no nice solution to this.
Can you not just fix the users (bmon?), after all it's ~only~ a compile time error. IFF_* flags are not provided by "route/link.h", users should include "net/if.h" or "linux/if.h" as they need it.

gustavoz:
On 04/17/2015 08:02 AM, Thomas Haller wrote:

> +#ifndef _LINUX_IF_H
> #include
> +#endif
> 
> _LINUX_IF_H is the include guard of "linux/if.h", so this check has
> basically no effect.
> 
> Maybe you wanted:
> +#ifndef _NET_IF_H
> #include
> +#endif
> 
> but that doesn't solve the issue from 50a7699
> <https://github.com/thom311/libnl/commit/50a76998ac36ace3716d3c979b352fac73cfc80a>
> in a nice way, because:
> 
>   * it relies on "net/if.h" to have include guard _NET_IF_H
>   * it still only works if you include "net/if.h" before "route/link.h".
> 
> I think there is no nice solution to this.
> Can you not just fix the users (bmon?), after all it's ~only~ a compile
> time error. IFF_* flags are not provided by "route/link.h", users should
> include "net/if.h" or "linux/if.h" as they need it.

github sent my second commit (testing) in the pull as well?
Odd, but not unexpected.
We could patch bmon but the documentation for libnl should be updated as
well (section 3.5.7) since users aren't told to include net/if.h there.
Regards.

thom311:
Which part of the documentation do you mean? I cannot find that part.

Could you quote some context around it? Thank you

thom311:
closing for inactivity. Please create new pull request if you think this should be solved in a particular way. Thank you and sorry.

PR #77, opened 2015-05-10
libnl: python swig api changes

arend:
Disabled debug macro in swig api to reduce the noise and exposed couple of functions.


thom311:
Looks good.

Merged as https://github.com/thom311/libnl/commit/eae6a6875c1f3b6f5c45b9bd39d5500622a239db


Thank you!

PR #78, opened 2015-06-08
fix path to cli linker version script for out-of-tree build

jfarrell:


thom311:
thank you. merged as https://github.com/thom311/libnl/commit/d754cb4543fddcd78f503aa69c5d8681568ec68d

PR #79, opened 2015-06-25
NUD_NOARP typo in trans_tbl?

toanju:
I would expect using rtnl_neigh_str2state("noarp") is translating to NUD_NOARP. Instead of "noarp" you have to use "norarp", which looks pretty much like a typo.

Adding "noarp" additionally might be an option as well...


thom311:
merged as https://github.com/thom311/libnl/commit/089a28212db9971a0aaa1952183de4f78dab3636

Thanks

PR #83, opened 2015-07-02
NFQUEUE: change integer types to more strict ones

socketpair:


lib/netfilter/queue_msg_obj.c
```diff
@@ -402,15 +402,18 @@ const uint8_t *nfnl_queue_msg_get_hwaddr(const struct nfnl_queue_msg *msg,
 	return msg->queue_msg_hwaddr;
 }
 
-int nfnl_queue_msg_set_payload(struct nfnl_queue_msg *msg, uint8_t *payload,
-			       int len)
+int nfnl_queue_msg_set_payload(struct nfnl_queue_msg *msg, const void *payload,
+			       unsigned int len)
 {
-	free(msg->queue_msg_payload);
```
socketpair: Forgot to add (I can split to two commits if you want) This fix case:
1. message has already some payload
2. malloc() failed

in that case `msg->queue_msg_payload_len` stay non-zero, `msg->queue_msg_payload`  become NULL.

It is fixed with my patch

thom311:
cherry-picked 3 patch as https://github.com/thom311/libnl/commit/e29c979e885ab3f16ab6b2b26a33bc079bb39c88

socketpair:
About changing API/ABI - Yes this will change header files, but in a way that WILL NOT break compiling applications and will not break aready compiled programs.

All changes does not change `sizeof()`s, just change signedness and constness. All that values are positive normally, so should not break anything (except things that was already was broken)

socketpair:
Unfortunatelly I close this PR, since n oone interested in that change :(

PR #86, opened 2015-07-09
Set ifi_change in link message

matben2:
The ifi_change field can be set with the mask of the flags that need to be changed as part of the link message to the kernel. This means only the specific flags that have been changed will be modified in the kernel, rather than the entire flags entry.


thom311:
I applied the patch to master as https://github.com/thom311/libnl/commit/34ccb7210f1238f89229c117dc3d28cea7ae00bb

I extended your patch to include a capability. With this a user could check whether the libnl version has this fix. I think that could be interesting, because this is a change in behavior.

Thank you

PR #87, opened 2015-08-05
lib/socket: add network namespace support to sockets

matben2:
It is possible to change the network namespace that a netlink socket
is listening in. This patch adds the function 'nl_socket_set_ns()'
to allow this to be configured.


thom311:
Just wondering, why did you withdraw the pull-request?

mattbennett2:
After seeing the build with the patch fail on jenkins I realised the SO_NSID option is not part of vanilla linux, rather an old VRF patch applied to the kernel I work with. Therefore I withdrew the pull-request.

PR #88, opened 2015-08-13
lib/nl: add missing header

joerg-krause:
CMSG_NXTHDR requires <linux/socket.h>. This fix a build error with the musl C library:

```sh
undefined reference to `__cmsg_nxthdr'
```


thom311:
merged as https://github.com/thom311/libnl/commit/d0406d3838d4e70f19dcac1747decd7dd9b7435c

Thank you

PR #89, opened 2015-09-19
enabled generation of Doxygen tagfile for libnl API reference

stv0g:
This enables other projects which are using Doxygen for documentation
to automatically create hyperlinks to libnl's API reference.

See: http://www.stack.nl/~dimitri/doxygen/manual/external.html


thom311:
Looks good.

Thank you!!

merged as 69851e24c7db9474316c5d97caf5e645fac69816

Issue #90, opened 2015-10-19
vlan_put_attrs missing IFLA_VLAN_PROTOCOL 


ElSzymono:
The vlan_put_attrs function in lib/route/link/vlan.c is missing:
```
	if (vi->vi_mask & VLAN_HAS_ID)
		NLA_PUT_U16(msg, IFLA_VLAN_ID, vi->vi_vlan_id);

+	if (vi->vi_mask & VLAN_HAS_PROTOCOL)
+		NLA_PUT_U16(msg, IFLA_VLAN_PROTOCOL, vi->vi_protocol);

	if (vi->vi_mask & VLAN_HAS_FLAGS) {
		struct ifla_vlan_flags flags = {
			.flags = vi->vi_flags,
			.mask = vi->vi_flags_mask,
		};
```
Because of this the protocol value set by rtnl_link_vlan_set_protocol is not passed to the kernel and only 802.1Q interfaces can be created.

thom311:
Fixed. See https://github.com/thom311/libnl/pull/91

PR #91, opened 2015-10-20
IFLA_VLAN_PROTOCOL added to vlan_put_attrs

ElSzymono:
This is a pull request regarding issue  #90 vlan_put_attrs missing IFLA_VLAN_PROTOCOL. This fix allows libnl to create not only 802.1Q VLAN interfaces but 802.1ad as well.


thom311:
applied as 0aca70083e51d97d0102c73852d31625f75c942b.

Thank you for the patch


I was considering to fix the missing ntohs(), but I'd rather not change behavior at this point. Thus I added a commit to document the odd behavior.

PR #92, opened 2015-11-19
route/link corrected array size

toanju:
Using now the macro for inet instead of inet6


thom311:
Merged as https://github.com/thom311/libnl/commit/acdeaf498d53d04ef4ca0e744b6c47e07e76c121

Thanks!!

PR #94, opened 2015-12-16
route/link/bridge: fixed return type

toanju:
Fixes: 6555b0d


thom311:
Thanks. Didn't get a compiler warning for this...

guess we should enable more warnings (and fix the false warnings)

PR #96, opened 2016-04-29
support neighbour flag NTF_SELF

jonasj76:
Adds support for `NTF_SELF` flag, e.g when using `rtnl_neigh_flags2str()`.


thom311:
Thanks. Merged as 648338a2170854952e605175af141ee27cc8dd2b

PR #97, opened 2016-04-29
add support for macsec

qsn:
For now this is only the part for creating macsec devices and configuring device attributes over RTNL-link.
The rest of the attributes of macsec devices is configured over genl, and is not part of this PR.

macsec devices are part of kernel 4.6-rc1.


libnl-route-3.sym
```diff
@@ -396,6 +396,33 @@ global:
 	rtnl_link_is_veth;
 	rtnl_link_is_vlan;
 	rtnl_link_is_vxlan;
+	rtnl_link_macsec_alloc;
```
thom311: These parts must be below in the libnl_3_2_28 section.

lib/route/link/macsec.c
```diff
@@ -0,0 +1,571 @@
...
+	struct macsec_info *info = link->l_info;
+
+	IS_MACSEC_LINK_ASSERT(link);
+
+	info->window = window;
+
+	return 0;
+}
+
+uint32_t rtnl_link_macsec_get_window(struct rtnl_link *link)
```
thom311: I know, there are many getters that follow this pattern. But this gives no way to return an error.

I'd prefer something like
  int rtnl_link_macsec_get_window(struct rtnl_link *link, uint32_t *window);
Like in commit 207dc0352779af46121a83fb4758cdd3155d9001

Also, usually rtnl objects have a mask field so you know which attributes are present or which not. I know it's cumbersome to add, but could we add that?

qsn:
Thanks for the comments, Thomas.
I will add what you suggested. The mask would also fix the clang build error.

thom311:
part of this was merged as https://github.com/thom311/libnl/pull/99

thom311:
for the record: the other patches are now on https://github.com/thom311/libnl/pull/100

PR #98, opened 2016-05-01
Add function to look up neighbour (FDB) by ifindex, mac address and vlan id

jonasj76:
The` rtnl_neigh_get()` function can not be used to look up a FDB entry in the
neigh cache. This is due to that the function searches among destination
addresses (`NDA_DST`) and not among link layer addresses (`NDA_LLADDR`), which is
used by FDB entries. A FDB entry can also exist in several VLANss, so a VLAN id
parameter is also needed to find a unique entry.
This commit adds a function, `rtnl_neigh_get_by_vlan()` which searches the neigh
cache for a specific neighbour (FDB) entry by interface index, link layer
address and VLAN id.


thom311:
Merged to master: https://github.com/thom311/libnl/commit/3bf503d30c271822158414f63eed620afc9b10cd


Thanks!

PR #99, opened 2016-06-24
Better flags handling in ->io_compare

qsn:
In order to match links using only attributes that are actually set in the filter object for cache lookups, we need to pass flags (LOOSE_COMPARISON) down to the io_compare op.
Also update vxlan's implementation of this op to handle LOOSE_COMPARISON.


thom311:
merged as https://github.com/thom311/libnl/commit/f087f3b50d82696d1c23b825ab87b46f0f167a50

Thanks!

PR #100, opened 2016-06-25
macsec device creation

qsn:
This adds support for creating and configuring macsec devices over rtnl.


thom311:
Looks very good. Merged https://github.com/thom311/libnl/commit/9b7c28ebd2686584fa9e5d35845f86d3b38e1fa9


I was only wondering about doing strict checking of the set-arguments like in:
```
+int rtnl_link_macsec_set_encrypt(struct rtnl_link *link, uint8_t encrypt)
+{
+    struct macsec_info *info = link->l_info;
+
+    IS_MACSEC_LINK_ASSERT(link);
+
+    if (encrypt > 1)
+         return -NLE_INVAL;
```
maybe we should just accept bogus values and let kernel reject invalid values.

But no strong opinion on that (and it's not handled consistently in the library). So, fine with me.


Thanks!

PR #101, opened 2016-07-13
Update neighbour discovery enum to match the linux definition

swstephenson:
A libnl enumeration has become out of sync with the one
used by the kernel.


thom311:
All the headers in include/linux-private/linux directory were copied at one point from kernel sources. It is to be expected (and OK) that they get out of sync.

Another issue is, that all these local headers are not taken from the same kernel release, but merged together from different versions. That is a bit ugly, but usually no problem because those headers are supposed to be backward compatible.

It's not clear why you want to extend that enum with new values, without using them. Ok, you bump NDA_MAX, which has effects on `neigh_policy[NDA_MAX+1]` ... but that doesn't seem to make any actual difference during  `nlmsg_parse(n, sizeof(*nm), tb, NDA_MAX, neigh_policy);` -- actually, it's a tiny bit worse, because neigh_policy grows by a few unused bytes and parsing takes a bit longer (with the same result).

I'm inclined to reject this patch until the new values are actually used. Please feel invited to re-open if you can give a reason why it should be applied (It doesn't have to be a very strong argument).

Thanks!

PR #102, opened 2016-07-20
Do not include linux/in6.h

jer-gentoo:
In file included from ../include/linux-private/linux/if_bridge.h:18:0,
                 from route/link/bridge.c:26:
.../usr/include/linux/in6.h:30:8: error: redefinition of ‘s
truct in6_addr’
 struct in6_addr {

### DIFF
--- a/include/linux-private/linux/if_bridge.h
+++ b/include/linux-private/linux/if_bridge.h
@@ -15,7 +15,6 @@
 
 #include <linux/types.h>
 #include <linux/if_ether.h>
-#include <linux/in6.h>
 
 #define SYSFS_BRIDGE_ATTR	"bridge"
 #define SYSFS_BRIDGE_FDB	"brforward"


thom311:
The header files in `include/linux-private` were originally copied over from linux kernel sources.

Preferably, they stay similar to the kernel headers and only updated by copying over new things from the kernel sources. Of course, over time the kernel headers change (so from time to time we re-import parts of the header files). And thus, the header files are not a snapshot of one particular kernel commit but merged together from different kernel versions.
In practice that isn't too much of a problem, because these header files are supposed to be backward compatible and not to break API. So, mostly it works, unless it doesn't then we fix it somehow -- but always keeping libnl3's copy as similar as possible to the one from kernel.


But your patch changes the header file with something that is not in the kernel sources (that is, the original if_bridge.h header does include <linux/in6.h>). That is wrong, and a different solution is needed there.


I cannot reproduce this build failure, can you show more failure messages?

thom311:
(closing for now, to mark this handled for myself. Please reopen or open a new pull-request. Thanks!)

PR #105, opened 2016-08-04
route/addr: address attributes based on object

toanju:
addr_id_attrs_get returned a fixed set of attributes for AF_INET. This
leads to an invalid cache in case the default cache manager is used.

The error was cause by nl_object_identical, which checkes the ce_mask
of an object against the req_attrs. For route/addr objects the ce_mask
may contain the ADDR_ATTR_PEER, but the addr_id_attrs_get always
includes this attribute. Thus nl_object_identical fails always in case
no peer exists, which is the default for local addresses.

fixes 62f0fdb9

### DIFF
--- a/lib/route/addr.c
+++ b/lib/route/addr.c
@@ -467,12 +467,15 @@ static void addr_dump_stats(struct nl_object *obj, struct nl_dump_params *p)
 static uint32_t addr_id_attrs_get(struct nl_object *obj)
 {
 	struct rtnl_addr *addr = (struct rtnl_addr *)obj;
+	uint32_t rv;
 
 	switch (addr->a_family) {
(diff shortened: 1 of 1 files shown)


thom311:
ow, this is a bad regression :( ... a good reason to do 3.2.29 sooner then expected.

merged: https://github.com/thom311/libnl/commit/5e17c54a7536caca42bacef4a01cf33ce131de0d


Thanks!

afett:
Do you still consider doing a "sooner" release to fix this, and if so... when would that be?

thom311:
I guess "sooner" became not soon...
There are still unmerged patches on the mailing list. Those should go in before a 3.2.29 release.

I plan to do that before the end of this year.

PR #107, opened 2016-08-18
compat: enable building on older systems (e.g., RHEL 6)

jsquyres:
This type is not present on older systems (e.g., RHEL 6), and libnl3 will not built without it.  So put a check in configure to ensure that this type is available.

Signed-off-by: Jeff Squyres <jsquyres@cisco.com>

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -153,6 +153,7 @@ noinst_HEADERS = \
 	linux-private/linux/netlink.h \
 	linux-private/linux/pkt_cls.h \
 	linux-private/linux/sock_diag.h \
+	linux-private/linux/socket.h \
 	linux-private/linux/tc_act/tc_mirred.h \
 	linux-private/linux/tc_act/tc_skbedit.h \
 	linux-private/linux/pkt_sched.h \
--- a/include/linux-private/linux/socket.h
+++ b/include/linux-private/linux/socket.h
@@ -1,0 +1,21 @@
+#ifndef _LINUX_SOCKET_H
+#define _LINUX_SOCKET_H
+
+/*
+ * Desired design of maximum size and alignment (see RFC2553)
+ */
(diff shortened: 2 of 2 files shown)


thom311:
Hi,

we would like to also build on old kernels... aside that apparently nobody is really doing that (and worse: testing).

libnl3 copies kernel headers to `include/linux-private` precisely to build with older kernel. In this case, it seems the right fix is to copy `linux/socket.h` to provide the type.
-- usually after something like `make headers_install INSTALL_HDR_PATH=./TMP`, which preprocesses the header file.

Would you be willing to send a patch for that? (plus testing)

Thank you

jsquyres:
@thom311 Sure, I'd be happy to.

...but I'm unsure what you're referring to with `make headers_install ...`.  That does not appear to be a `make` target in libnl3.  Are you referring to a target in some other package?

thom311:
Cool thanks, you did all correctly :)
I'll apply it shortly.



I meant that usually the uapi header files from within the kernel source tree should be preprocessed before being used in the user-space.
This is what  `make headers_install INSTALL_HDR_PATH=./TMP` does. Effectively, it just changes

    #ifndef _UAPI_LINUX_SOCKET_H
    #define _UAPI_LINUX_SOCKET_H

to

    #ifndef _LINUX_SOCKET_H
    #define _LINUX_SOCKET_H

jsquyres:
New commit pushed up with `/usr/include/linux/socket.h` from a RHEL 7 install.

It compiles, builds, and passes `make check` on both a RHEL 6 and RHEL 7 machine.

(...and you noticed it before I pushed the "comment" button to publish this comment 😄 )

thom311:
merged: https://github.com/thom311/libnl/commit/746bbba3e55206c67808d3243aad740abc92ab3f

Thanks!!

PR #108, opened 2016-09-06
macsec: minor bugfix

dcaratti:
this  pull request contains two small fixes for macsec:
* limit the maximum ICV length (16 octets instead of 32)
* fix endianness of 'sci' parameter 

In addition, multiple implementations hof htonll() and ntohll() have been replaced with a single #define in include/netlink-private/utils.h


thom311:
merged: https://github.com/thom311/libnl/commit/ff7ca59783e8ad8bd40f9dce349bbe06b64b06fa

Thanks

PR #109, opened 2016-09-25
macvlan: add support for "source" mode

michael-dev:
This adds libnl support for new "source" mode.

Signed-off-by: Michael Braun <michael-dev@fami-braun.de>

### DIFF
--- a/include/netlink/route/link/macvlan.h
+++ b/include/netlink/route/link/macvlan.h
@@ -29,6 +29,9 @@ extern int		rtnl_link_macvlan_str2mode(const char *);
 extern char *		rtnl_link_macvlan_flags2str(int, char *, size_t);
 extern int		rtnl_link_macvlan_str2flags(const char *);
 
+extern char *		rtnl_link_macvlan_macmode2str(int, char *, size_t);
+extern int		rtnl_link_macvlan_str2macmode(const char *);
+
 extern int		rtnl_link_macvlan_set_mode(struct rtnl_link *,
 			                           uint32_t);
 extern uint32_t		rtnl_link_macvlan_get_mode(struct rtnl_link *);
@@ -38,6 +41,21 @@ extern int		rtnl_link_macvlan_set_flags(struct rtnl_link *,
 extern int		rtnl_link_macvlan_unset_flags(struct rtnl_link *,
 						   uint16_t);
 extern uint16_t		rtnl_link_macvlan_get_flags(struct rtnl_link *);
+
+extern int		rtnl_link_macvlan_set_macmode(struct rtnl_link *,
+						      uint32_t);
+extern int		rtnl_link_macvlan_get_macmode(struct rtnl_link *link,
+						      uint32_t *out_macmode);
+
+extern int		rtnl_link_macvlan_count_macaddr(struct rtnl_link *link,
+							uint32_t *out_count);
+extern int		rtnl_link_macvlan_get_macaddr(struct rtnl_link *link,
+						      unsigned int idx,
+						      const struct nl_addr **addr);
+extern int		rtnl_link_macvlan_add_macaddr(struct rtnl_link *link,
+						      struct nl_addr *addr);
+extern int		rtnl_link_macvlan_del_macaddr(struct rtnl_link *link,
+						      struct nl_addr *addr);
 
 #ifdef __cplusplus
 }
--- a/lib/route/link/macvlan.c
+++ b/lib/route/link/macvlan.c
@@ -38,12 +38,17 @@
 /** @cond SKIP */
 #define MACVLAN_HAS_MODE	(1<<0)
 #define MACVLAN_HAS_FLAGS	(1<<1)
+#define MACVLAN_HAS_MACCOUNT	(1<<2)
+#define MACVLAN_HAS_MACDATA	(1<<3)
 
 struct macvlan_info
 {
 	uint32_t		mvi_mode;
 	uint16_t		mvi_flags; // there currently is only one flag and kernel has no flags_mask yet
 	uint32_t		mvi_mask;
+	uint32_t		mvi_maccount;
+	uint32_t		mvi_macmode;
+	struct nl_addr		**mvi_macaddr;
 };
 
 /** @endcond */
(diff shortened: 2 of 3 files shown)


include/netlink/route/link/macvlan.h
```diff
@@ -29,6 +29,9 @@ extern int		rtnl_link_macvlan_str2mode(const char *);
 extern char *		rtnl_link_macvlan_flags2str(int, char *, size_t);
 extern int		rtnl_link_macvlan_str2flags(const char *);
 
+extern char *		rtnl_link_macvlan_macmode2str(int, char *, size_t);
+extern int		rtnl_link_macvlan_str2macmode(const char *);
```
thom311: symbols must be explicitly exposed in the ABI, that is: add them to libnl-route-3.sym.

include/netlink/route/link/macvlan.h
```diff
@@ -39,6 +42,17 @@ extern int		rtnl_link_macvlan_unset_flags(struct rtnl_link *,
...
+extern int		rtnl_link_macvlan_set_macmode(struct rtnl_link *,
+						      uint32_t);
+extern int		rtnl_link_macvlan_get_macmode(struct rtnl_link *);
+
+extern int		rtnl_link_macvlan_count_macaddr(struct rtnl_link *link);
+extern struct nl_addr *	rtnl_link_macvlan_get_macaddr(struct rtnl_link *link, int idx);
+extern int		rtnl_link_macvlan_add_macaddr(struct rtnl_link *link,
+						      struct nl_addr *addr);
+extern int		rtnl_link_macvlan_del_macaddr(struct rtnl_link *link,
+						      struct nl_addr *addr);
```
thom311: Add to libnl-route-3.sym, in the libnl_3_2_29 section.

lib/route/link/macvlan.c
```diff
@@ -96,20 +110,58 @@ static int macvlan_parse(struct rtnl_link *link, struct nlattr *data,
...
 
 static void macvlan_free(struct rtnl_link *link)
 {
-	free(link->l_info);
+	struct macvlan_info *mvi;
+	int i;
+
+	mvi = link->l_info;
+
+	if (mvi->mvi_mask & MACVLAN_HAS_MACDATA)
```
thom311: multi-line statements need a { }

lib/route/link/macvlan.c
```diff
@@ -121,12 +173,22 @@ static void macvlan_dump(struct rtnl_link *link, struct nl_dump_params *p)
 		rtnl_link_macvlan_flags2str(mvi->mvi_flags, buf, sizeof(buf));
 		nl_dump(p, "%s-flags %s", link->l_info_ops->io_name, buf);
 	}
+
+	if (mvi->mvi_mask & MACVLAN_HAS_MACCOUNT)
+		nl_dump(p, "macvlan-count %d", mvi->mvi_maccount);
+
+	if (mvi->mvi_mask & MACVLAN_HAS_MACDATA)
```
thom311: same here. Needs { }

lib/route/link/macvlan.c
```diff
@@ -155,9 +227,23 @@ static int macvlan_put_attrs(struct nl_msg *msg, struct rtnl_link *link)
...
+			goto nla_put_failure;
+
+		for (i = 0; i < mvi->mvi_maccount; i++) {
+			NLA_PUT_ADDR(msg, IFLA_MACVLAN_MACADDR,
+				     mvi->mvi_macaddr[i]);
+		}
+	}
 
 nla_put_failure:
+	if (datamac)
```
thom311: seems macvlan_put_attrs() should return -NLE_NOMEM in case of a nla_put_failure and not 0. Can you fix that too?

lib/route/link/macvlan.c
```diff
@@ -249,12 +335,23 @@ int rtnl_link_is_macvlan(struct rtnl_link *link)
...
 	struct macvlan_info *mvi = link->l_info;
+	int i;
 
 	IS_MACVLAN_LINK_ASSERT(link);
 
 	mvi->mvi_mode = mode;
 	mvi->mvi_mask |= MACVLAN_HAS_MODE;
 
+	if (mode != MACVLAN_MODE_SOURCE) {
+		if (mvi->mvi_mask & MACVLAN_HAS_MACDATA)
```
thom311: needs { }

lib/route/link/macvlan.c
```diff
@@ -277,6 +374,55 @@ uint32_t rtnl_link_macvlan_get_mode(struct rtnl_link *link)
...
+
+/**
+ * Get MACVLAN MACMODE
+ * @arg link		Link object
+ *
+ * Only for SOURCE mode.
+ *
+ * @return MACVLAN mac list modification mode, 0 if not set or a negative error code.
+ */
+int rtnl_link_macvlan_get_macmode(struct rtnl_link *link)
```
thom311: can we change the signature to

int rtnl_link_macvlan_get_macmode(struct rtnl_link *link, uint32_t *out_macmode);

libnl is unfortunately not consistent about this, but I think getters should return an error-code or success only. The actual return value should be passed separately. Especially, as uint32_t doesn't fit entirely in the "int" return value.

lib/route/link/macvlan.c
```diff
@@ -332,6 +478,166 @@ uint16_t rtnl_link_macvlan_get_flags(struct rtnl_link *link)
 	return mvi->mvi_flags;
 }
 
+/**
+ * Get number of MAC-Addr for MACVLAN device in source mode
+ * @arg link		Link object
+ *
+ * @return number of addresses on success or a negative error code.
+ */
+int rtnl_link_macvlan_count_macaddr(struct rtnl_link *link)
```
thom311: same here.

lib/route/link/macvlan.c
```diff
@@ -332,6 +478,166 @@ uint16_t rtnl_link_macvlan_get_flags(struct rtnl_link *link)
...
+	    !(mvi->mvi_mask & MACVLAN_HAS_MACDATA)) {
+		APPBUG("You need to read the list of macs from the macvlan "
+		       "device before.");
+		return NULL;
+	}
+
+	if (idx >= mvi->mvi_maccount || idx < 0)
+		return NULL;
+
+	return nl_addr_clone(mvi->mvi_macaddr[idx]);
```
thom311: maybe the getter shouldn't return a clone of the address? Can it not return a const-pointer?

lib/route/link/macvlan.c
```diff
@@ -332,6 +478,166 @@ uint16_t rtnl_link_macvlan_get_flags(struct rtnl_link *link)
...
+{
+	struct macvlan_info *mvi = link->l_info;
+
+	IS_MACVLAN_LINK_ASSERT(link);
+
+	if (!(mvi->mvi_mask & MACVLAN_HAS_MODE) ||
+	    (mvi->mvi_mode != MACVLAN_MODE_SOURCE)) {
+		APPBUG("Only macvlan source mode links can have a list of "
+		       "source mac addresses. Set mode \"source\" first.");
+		return NULL;
```
thom311: maybe rtnl_link_macvlan_get_macaddr() also should return an "int" error-code.

And maybe just get away with these assertions. If the link has the wrong mode, then it should return just return -NLE_INVAL.

michael-dev:
Thanks, should be fixed now.

thom311:
Hi Michael,

I merged the patch to master.

Note that I added a follow-up commit that changes some things. If you disagree with any of those, let's discuss it or just send a new PR to fix them. Thanks!!

PR #112, opened 2016-10-27
Add PPP support

jonasj76:
Add support for kernel PPP interfaces.


lib/route/link/ppp.c
```diff
@@ -0,0 +1,219 @@
...
+ * @route_doc{link_ppp, PPP Documentation}
+ * @{
+ */
+
+#include <netlink-private/netlink.h>
+#include <netlink/netlink.h>
+#include <netlink-private/route/link/api.h>
+
+/** @cond SKIP */
+#define PPP_HAS_FD		(1<<0)
```
thom311: could we name this PPP_ATTR_FD for consistency with other types?

lib/route/link/ppp.c
```diff
@@ -0,0 +1,219 @@
...
+
+struct ppp_info
+{
+	uint32_t		pi_fd;
+	uint32_t		pi_mask;
```
thom311: most other implementations name this field ce_mask. Could we stick to that?

lib/route/link/ppp.c
```diff
@@ -0,0 +1,219 @@
...
+	return 0;
+}
+
+/**
+ * Get PPP file descriptor
+ * @arg link		Link object
+ *
+ * @return PPP file descriptor, 0 if not set or a negative error code.
+ */
+uint32_t rtnl_link_ppp_get_mode(struct rtnl_link *link)
```
thom311: should be named rtnl_link_ppp_get_fd() for consistency.

Also, I know libnl3's API is quite inconsistent, but I would prefer as signature
    int rtnl_link_ppp_get_fd (struct rtnl_link *link, uint32 *out_fd);
to return errors separate from the value.

jonasj76:
Updated code after review.

Also:
- Renaming local `struct ppp_info*` variables from `pi` -> `info`.
- Fix type of file descriptor: `uint32_t` -> `int32_t`.

thom311:
merged as https://github.com/thom311/libnl/commit/306695fb07a39a0e76b361ce62f5f38c28cf4f79 with one minor fix at the end.

-- also, during rebase, there were some conflicts that I resolved, so the patches are not 100 % original...


thanks!!

PR #113, opened 2016-10-28
xfrm: fix xfrm security context management

spellingmistake:
The data structure of choice when adding/processing a security context
for xfrm is struct xfrm(nl)_user_sec_ctx. The previous code did however
use the (also exported) struct xfrm(nl)_sec_ctx. While sizeof(struct
xfrm(nl)_*sec_ctx) yields the same result, the interpretation of one of
the data structures as the other one messes up the contents.
With this fix, the wrong data structure has been replaced with the
correct one. Also -- since the size of the context string is not known
-- one can now call xfrmnl_sa_get_sec_ctx with ctx_str being NULL, thus
retrieving the length of the context string.
A new capability has been introduced, to test whether libnl3 supports
the modified semantics of this function.

Signed-off-by: Thomas Egerer thomas.egerer@secunet.com

### DIFF
--- a/include/netlink/utils.h
+++ b/include/netlink/utils.h
@@ -231,6 +231,13 @@ enum {
 	NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR = 21,
 #define NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR
 
+	/**
+	 * Support omitting @ctx_str argument to xfrmnl_sa_get_sec_ctx() to check
+	 * for required buffer size for context string.
+	 */
+	NL_CAPABILITY_XFRM_SEC_CTX_LEN = 22,
+#define NL_CAPABILITY_XFRM_SEC_CTX_LEN NL_CAPABILITY_XFRM_SEC_CTX_LEN
+
 	__NL_CAPABILITY_MAX,
 	NL_CAPABILITY_MAX = (__NL_CAPABILITY_MAX - 1),
 #define NL_CAPABILITY_MAX NL_CAPABILITY_MAX
--- a/lib/xfrm/sa.c
+++ b/lib/xfrm/sa.c
@@ -191,7 +191,7 @@ static int xfrm_sa_clone(struct nl_object *_dst, struct nl_object *_src)
 
 	if (src->sec_ctx)
 	{
-		len = sizeof (struct xfrmnl_sec_ctx) + src->sec_ctx->ctx_len;
+		len = sizeof (*src->sec_ctx) + src->sec_ctx->ctx_len;
 		if ((dst->sec_ctx = calloc (1, len)) == NULL)
 			return -NLE_NOMEM;
 		memcpy ((void *)dst->sec_ctx, (void *)src->sec_ctx, len);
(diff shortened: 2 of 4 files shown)


lib/xfrm/sa.c
```diff
@@ -191,7 +191,7 @@ static int xfrm_sa_clone(struct nl_object *_dst, struct nl_object *_src)
 
 	if (src->sec_ctx)
 	{
-		len = sizeof (struct xfrmnl_sec_ctx) + src->sec_ctx->ctx_len;
+		len = sizeof (src->sec_ctx) + src->sec_ctx->ctx_len;
```
thom311: sizeof (*src->sec_ctx) ?

lib/xfrm/sa.c
```diff
@@ -1938,36 +1937,72 @@ int xfrmnl_sa_set_mark (struct xfrmnl_sa* sa, unsigned int value, unsigned int m
...
+ * @arg doi             An optional output value for the security context domain of interpretation.
+ * @arg alg             An optional output value for the security context algorithm.
+ * @arg len             An optional output value for the security context length, including the
+ *                      terminating null byte ('\0').
+ * @arg sid             Unused parameter.
+ * @arg ctx_str         An optional buffer large enough for the security context string. It must
+ *                      contain at least @len bytes.
+ *
+ * Warning: you must ensure that @ctx_str is large enough. If you don't know the length before-hand,
+ * call xfrmnl_sa_get_sec_ctx() without @ctx_str argument to query only the required buffer size.
```
thom311: maybe comment that older versions of libnl didn't have the arguments as optional and about NL_CAPABILITY_SEC_CTX_LEN?

include/netlink/utils.h
```diff
@@ -231,6 +231,13 @@ enum {
 	NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR = 21,
 #define NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR NL_CAPABILITY_NL_ADDR_FILL_SOCKADDR
 
+	/**
+	 * Support omitting @ctx_str argument to xfrmnl_sa_get_sec_ctx() to check
+	 * for required buffer size for context string.
+	 */
+	NL_CAPABILITY_SEC_CTX_LEN = 22,
```
thom311: maybe name it NL_CAPABILITY_XFRMNL_SEC_CTX_LEN?

lib/xfrm/sa.c
spellingmistake: You're right, good catch. This one slipped through my test program on a 64 bit machine, since sizeof(uintptr_t) is equal to sizeof(struct xfrmnl_user_sec_ctx).

include/netlink/utils.h
spellingmistake: Makes sense. However, since the SA_KEY_SIZE macro is has a NL_CAPIBILITY_XFRM_ prefix -- without the NL, I called the macro  NL_CAPABILITY_XFRM_SEC_CTX_LEN. Looks better to me. But feel free to change anything you don't like.

spellingmistake:
Thomas, I'm not familiar with github to the extend how to deal with my modifications of the pull-request. Naturally I would try and update this pull request, but as it does not seem possible, I'll close this one and make another pull request, with all the requested changes.

thom311:
Usually you would just force-push the updated branch to your github repo.
Anyway, close + open-new is fine too. thanks!!

Follow up: https://github.com/thom311/libnl/pull/114

PR #114, opened 2016-11-06
Fix xfrm sec ctx

spellingmistake:
Second attempt, adding the upated doc for the xfrmnl_sa_get_* functions as well as the suggested modifications/fixes.

### DIFF
--- a/include/netlink-private/types.h
+++ b/include/netlink-private/types.h
@@ -1155,6 +1155,7 @@ struct xfrmnl_encap_tmpl {
 	struct nl_addr* encap_oa;
 };
 
+/* kernel space structure, don't use to retrieve/pass sec ctx attribute! */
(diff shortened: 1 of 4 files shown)


thom311:
It looks good to me now, didn't test it though. I hope you tested it :)

Thanks, merged: https://github.com/thom311/libnl/commit/d9840f648cf724e0d43dfe33e5a24f026f7fa52e

This is a rather ugly bug, because although sizeof(xfrmnl_sec_ctx) is the same as sizeof(xfrmnl_user_sec_ctx), the ctx_len is located somewhere differen.
I don't think somebody can reasonably use this API prior this fix, thus an `assert(nl_get_capability(NL_CAPABILITY_XFRM_SEC_CTX_LEN))` might be advised.

PR #115, opened 2016-11-11
Add MULTIPATH to route attribute list

anantdeepak:
When comparing route objects, we need to include multipath attribute so routes which have the same prefixes via different nextHop interfaces can
be differentiated. Example: Multicast routes: ff00::/8 via eth0 and
ff00::/8 via eht1

### DIFF
--- a/lib/route/route_obj.c
+++ b/lib/route/route_obj.c
@@ -1310,7 +1310,7 @@ struct nl_object_ops route_obj_ops = {
 	.oo_attrs2str		= route_attrs2str,
 	.oo_id_attrs		= (ROUTE_ATTR_FAMILY | ROUTE_ATTR_TOS |
 				   ROUTE_ATTR_TABLE | ROUTE_ATTR_DST |
-				   ROUTE_ATTR_PRIO),
+				   ROUTE_ATTR_PRIO | ROUTE_ATTR_MULTIPATH),
 };
 /** @endcond */
 


thom311:
Hi @anantdeepak,

this patch is not the correct fix because the multipath attribute is not part of the ID. Well, in a sense it is, because for the kernel *every* route property is part of the ID. See for example this bug: https://bugzilla.redhat.com/show_bug.cgi?id=1337855

One solution with the current kernel API was discussed here: https://lists.infradead.org/pipermail/libnl/2016-June/002142.html
But these patches are note merged, because they seem complicated and (obviously) change behavior. I think the patchset is right and should be merged... but the solution there is quite ugly, so I would hope that kernel would instead improve the netlink API...


I close this PR for now, feel free to reply and re-open. Thank you.

anantdeepak:
@thom311 

Thanks for the explanation.... 
For now, we are using this patch internally to work around our problem (specific to same prefix on different dev) 

I hope this gets resolved soon. 

I wonder, if until a more elegant solution (a 64 bit unique ID for route record) is implemented, having libnl specify all known attributes as a stop-gap ? Admitted, it will be a weak implementation which *may* break if kernel attributes change in the future, but it is better than a known broken implementation...

What do you think ?

thom311:
Maybe. There is also https://bugzilla.redhat.com/show_bug.cgi?id=1337860 . That wouldn't matter while doing a full dump, but it matters if you have a filled cache and want to update it based on RTM_NEW_ROUTE/RTM_DEL_ROUTE notifications.

https://lists.infradead.org/pipermail/libnl/2016-June/002142.html gets this right (I think), this should be merged.

PR #116, opened 2016-11-15
lib/route: keep link stats minlen compatible with kernel < 4.6

tklauser:
Since linux/if_link.h was synced in libnl commit d533736e2258 ("vxlan:
add support for additional VXLAN attributes."), the minlen check in
validate_nla() fails on kernels < 4.6 - or more precisely, on kernels
not containing kernel commit 6e7333d315a7 ("net: add rx_nohandler stat
counter"). The check fails because rtnl_link_stats and rtnl_link_stats64
sent by the kernel don't have the rx_nohandler member yet, but libnl
expects them to have it.

Account for this by decreasing the minlen by one member and add an
explanatory comment.

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -313,6 +313,11 @@ static int link_clone(struct nl_object *_dst, struct nl_object *_src)
 	return 0;
 }
 
+/* struct rtnl_link_stats doesn't have rx_nohandler in kernel versions < 4.6 */
+#define IFLA_STATS_MINLEN (sizeof(struct rtnl_link_stats) - sizeof(__u32))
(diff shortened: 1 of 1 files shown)


thom311:
Hi Tobias,

merged as commit 51a51c3a1c7285b038ab7d5d45a3b5c9ade199ea.

Note that I pushed a follow-up commit that changes your patch to something else :)
Please take a look (and speak up if you dislike it).

thanks for the patch.

tklauser:
Hi Thomas,

Your solution looks much better and robust against future stat counter additions and I don't dislike it at all :) 

Thanks a lot for following up!

PR #117, opened 2016-11-29
Wire up rx_nohandler in rtnl link stats

tklauser:
Hi Thomas,

As a follow-up to #116, here's a PR to add the newly introduced rx_nohandler counter to libnl.

Tested on kernel 4.6 and 4.8 with nl-link-stats (after fixing the off-by-one issue with the stats maximum used to iterate through the stats).

Thanks
Tobias


thom311:
Thanks Tobias,

merged your patches + two follow-ups.

PR #119, opened 2016-12-06
Add support for more rtnl attributes

tklauser:
Add support for the following rtnl attributes:

* IFLA_CARRIER_CHANGES
* IFLA_PHYS_PORT_NAME
* IFLA_PHYS_SWITCH_ID
* IFLA_GSO_MAX_SEGS

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -69,6 +69,8 @@
 #define LINK_ATTR_CARRIER_CHANGES	((uint64_t) 1 << 34)
 #define LINK_ATTR_PHYS_PORT_NAME	((uint64_t) 1 << 35)
 #define LINK_ATTR_PHYS_SWITCH_ID	((uint64_t) 1 << 36)
+#define LINK_ATTR_GSO_MAX_SEGS		((uint64_t) 1 << 37)
+#define LINK_ATTR_GSO_MAX_SIZE		((uint64_t) 1 << 38)
 
 static struct nl_cache_ops rtnl_link_ops;
 static struct nl_object_ops link_obj_ops;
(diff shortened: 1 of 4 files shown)


libnl-route-3.sym
```diff
@@ -286,6 +286,7 @@ global:
 	rtnl_link_get_broadcast;
 	rtnl_link_get_by_name;
 	rtnl_link_get_carrier;
+	rtnl_link_get_carrier_changes;
```
thom311: a section in the linker-version script must never be modified once a new API is released. That means, new symbols must always be added to the section for the next upcoming release, that is `libnl_3_2_29` at the moment. I mean, new symbols must be added at last section in this file.

lib/route/link.c
```diff
@@ -2273,6 +2284,17 @@ uint8_t rtnl_link_get_carrier(struct rtnl_link *link)
 }
 
 /**
+ * Return carrier on/off changes of link object
+ * @arg link		Link object
+ *
+ * @return Carrier changes.
+ */
+uint32_t rtnl_link_get_carrier_changes(struct rtnl_link *link)
```
thom311: libnl is not very consistent about this, but I think that in most cases getters should return an `int` value indicating success (negative values meaning error, `NLE_*`). Something like https://github.com/thom311/libnl/blob/993a62d4501a9b40461d9f1276a63246268b4a3a/lib/route/link/ppp.c#L197

Not sure about getters that return pointers like `rtnl_link_get_phys_switch_id`. But for getters that return integers, I think we shouldn't mix an error result with the value.

libnl-route-3.sym
tklauser: Ah yes, didn't realize that. Will change accordingly (also the other .sym entries)

lib/route/link.c
tklauser: Most of the other `IFLA_*` "base" link attributes use these simple getters AFAICS, so I thought I stick to the pattern to make usage consistent.

But if you prefer the `rtnl_link_ppp_get_fd` style, I'll update the integer getters accordingly.

thom311: It's not clear what's best, and consistency should be aimed for... but if the return value is uint32_t, there is no space to encode an error. Or even to encode whether the attribute is present -- which might be relevant information. I think a `int` error code is better.

thom311:
Merged: https://github.com/thom311/libnl/commit/44880b68d847894cc71ec0bc5f774c2ccd88a0bf

Thank you!

PR #120, opened 2016-12-06
link: fix documentation for rtnl_link_get_carrier_changes

tklauser:
Sorry, I was a bit too fast to push out after fixing up the getters in #119. So here's a small fix for the doxygen comment of `rtnl_link_get_carrier_changes`. The doxygen comments for the other newly introduced getters should be correct already.


thom311:
thanks!! merged

PR #121, opened 2017-01-01
Remove unneeded include in private headers

jer-gentoo:
Including linux/in6.h breaks with 3.4 (and earlier?) headers:

In file included from ../include/linux-private/linux/if_bridge.h:18:0, from route/link/bridge.c:26: /usr/include/linux/in6.h:30:8: error: redefinition of 'struct in6_addr'

Also reported at https://bugs.gentoo.org/show_bug.cgi?id=604190

### DIFF
--- a/include/linux-private/linux/if_bridge.h
+++ b/include/linux-private/linux/if_bridge.h
@@ -15,7 +15,6 @@
 
 #include <linux/types.h>
 #include <linux/if_ether.h>
-#include <linux/in6.h>
 
 #define SYSFS_BRIDGE_ATTR	"bridge"
 #define SYSFS_BRIDGE_FDB	"brforward"


thom311:
when building libnl3, the library brings it's own copy of kernel headers in include/linux-private.
It is not supposed to include kernel headers from /usr/include/linux, and it doesn't do that for me.

See that there is a file "include/linux-private/linux/in6.h" should should be used instead.

Also, the files in include/linux-private are copied from the original kernel headers. They are not supposed to be modified to keep the copy close to the original.

I think you are doing something odd when building libnl(?)

jer-gentoo:
As https://bugs.gentoo.org/show_bug.cgi?id=604190#c0 shows, when building against 3.4, the headers conflict, and removing this single include line fixes the issue. I have carried this same patch in a kernel 3.4 limited buildroot (https://buildroot.org/) tree for half a year now.

thom311:
https://bugs.gentoo.org/show_bug.cgi?id=604190#c0 has:

     In file included from ../include/linux-private/linux/if_bridge.h:18:0,
                     from route/link/bridge.c:26:
     /usr/include/linux/in6.h:30:8: error: redefinition of 'struct in6_addr'

it is not clear (to me) why "linux/in6.h" would be included from `/usr/include/linux/in6.h`, instead of ../include/linux-private/linux/in.6. That is wrong and that is a first issue.

Note that you should not build against 3.4 kernel headers. Instead, you build against the kernel headers included in libnl3 in "include/linux-private" (as they are).


Then the actual build failure is due to a conflict in `struct in6_addr`, which is either provided by netinet/in.h or linux/in.6. I thought such conflicts have long been fixed. Is this with recent glibc headers?

jer-gentoo:
I don't understand. You don't want the compiler to use the "system" `in6.h` but you don't include `in6.h` in `include/linux-private` either? Obviously my patch is wrong if `in6.h` was omitted by mistake, in which case it should be added in the distribution.

thom311:
Ah! I see, in6.h is present in the git-tree but not disted in the tarball.

Your https://github.com/thom311/libnl/pull/127 seems right... follow up there.

PR #122, opened 2017-01-13
route/tc: Remove unused function tca_set_kind()

tklauser:
The public prototype and the last internal user of the function were
removed in commit 8eb5b5532e ("Unified TC API") and it was unexported in
commit 4280dfb85d ("build: don't export internal symbols"), so it is
safe to remove it.


thom311:
merged

Thanks!!

PR #127, opened 2017-01-17
in6.h: Distribute and drop libc-compat include

jer-gentoo:


jer-gentoo:
Oh, I see you have a `libc-compat.h` that isn't distributed either, so maybe instead of removing the include, you should distribute that as well?

thom311:
> Oh, I see you have a libc-compat.h that isn't distributed either, so maybe instead of removing the include, you should distribute that as well?

ah yes. We don't want any linux headers from the system, only those from `include/linux-private`.

So, `include/linux-private` should be self-contained, and the files in the git-tree should also be dist-ed in `Makefile.am` -- as you do for `in6.h`.

Could you also add `libc-compat.h` to Makefile.am (and whatever you see is missing). Thanks!


Does that fix the build issue you had with https://github.com/thom311/libnl/pull/121 ?

thom311:
`in.h` is also not distributed...

PR #130, opened 2017-01-20
Distribute in.h in6.h libc-compat.h

jer-gentoo:
I also sorted the entries alphabetically.


thom311:
merged as commit https://github.com/thom311/libnl/commit/6c7f42150034b9e1d8d9d7defc37a72f717ec36a

Thanks!!

PR #133, opened 2017-02-06
Little fixes in creation of security policy messages

HolyShitMan:
I stumbled over some "kind-of-bugs" in _libnl/lib/xfrm/sp.c_ which hindered me to create and delete xfrm policies.
1. The check, whether the policy index and direction set (add and delete message) is to restrictive. Policies can be identified either with index and direction or with direction, selector and security context ,where "not set" is valid too (see kernel/net/xfrm/xfrm_user.c -  `xfrm_get_policy(...)` )
2. In xfrmnl_sp_set_sec_ctx the length values were calculated kind of strange. Firstly both parameters should be different and secondly, the length can be calculated directly from length-of-string and `sizeof(struct  xfrmnl_user_sec_ctx)`.
I don't think that anybody allready used `xfrmnl_sp_set_sec_ctx()` because the old version doesn't work, but i decided not to change the interface and only use a compiler attribute to suppress warnings for unused parameter `len`
3. Constructing a delete-message without index, but with selector and security-context wasn't possible. I added the necessary code.

### DIFF
--- a/lib/xfrm/sp.c
+++ b/lib/xfrm/sp.c
@@ -765,8 +765,7 @@ static int build_xfrm_sp_message(struct xfrmnl_sp *tmpl, int cmd, int flags, str
 	uint32_t                    len;
 	struct nl_addr*             addr;
 
-	if (!(tmpl->ce_mask & XFRM_SP_ATTR_INDEX) ||
-	    !(tmpl->ce_mask & XFRM_SP_ATTR_DIR))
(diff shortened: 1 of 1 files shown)


thom311:
Branch merged, with two additional commits (commit fc21d073d97dc186e5a005c0182a72b707863b2c).

Closing, please re-open (or open a new PR) if there is another issue.

thanks!!

PR #135, opened 2017-03-12
lib, src: fix build with uclibc

baruchsiach:
uClibc-ng C library does not implement strerror_l(). Fallback to plain
strerror() to fix the build.

Signed-off-by: Baruch Siach <baruch@tkos.co.il>

### DIFF
--- a/lib/utils.c
+++ b/lib/utils.c
@@ -121,6 +121,10 @@ int __nl_read_num_str_file(const char *path, int (*cb)(long, const char *))
 	return 0;
 }
 
+#ifdef __UCLIBC__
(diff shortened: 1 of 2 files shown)


baruchsiach:
Unfortunately, this does not fix the whole problem. uClibc-ng can be configured without locale support, causing more build failures. I'll try a better fix.

thom311:
2 days ago, a similar patch was sent do the mailing list:
http://lists.infradead.org/pipermail/libnl/2017-March/002301.html

I merged it as https://github.com/thom311/libnl/commit/e15966ac7f3b43df2acf869f98089762807d0568

PR #136, opened 2017-03-13
route/link: add accessor API for IPv6 flags

matra01:
Add functions to access the IPv6 specific flags of a link object.
Also the functions for IPv6 link flags translation are now exported, similar
to the non IPv6 specific translation functions.

### DIFF
--- a/include/netlink/route/link/inet6.h
+++ b/include/netlink/route/link/inet6.h
@@ -36,6 +36,16 @@ extern int		rtnl_link_inet6_get_addr_gen_mode(struct rtnl_link *,
 extern int		rtnl_link_inet6_set_addr_gen_mode(struct rtnl_link *,
 							  uint8_t);
 
+extern int		rtnl_link_inet6_get_flags(struct rtnl_link *,
+							  uint32_t *);
+
+extern int		rtnl_link_inet6_set_flags(struct rtnl_link *,
+							  uint32_t);
+
+/* Link Flags Translations */
+extern char *	rtnl_link_inet6_flags2str(int, char *, size_t);
+extern int		rtnl_link_inet6_str2flags(const char *);
+
 #ifdef __cplusplus
 }
 #endif
--- a/libnl-route-3.sym
+++ b/libnl-route-3.sym
@@ -1032,3 +1032,10 @@ global:
 	rtnl_u32_set_selector;
 } libnl_3_2_28;
 
+libnl_3_4 {
+global:
+	rtnl_link_inet6_flags2str;
+	rtnl_link_inet6_str2flags;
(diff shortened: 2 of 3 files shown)


libnl-route-3.sym
```diff
@@ -843,10 +843,14 @@ global:
 	rtnl_class_hfsc_set_usc;
 	rtnl_link_inet6_addrgenmode2str;
 	rtnl_link_inet6_get_addr_gen_mode;
+	rtnl_link_inet6_get_flags;
```
thom311: a symbol version section in the linker script cannot ever be modified after a release is out.

Actually, 3.3.0-rc1 is already out. I would prefer not to add new API after the release-candidate. Can we target 3.4.0 (with linker section "libnl_3_4")?

include/netlink/route/link/inet6.h
```diff
@@ -36,6 +36,16 @@ extern int		rtnl_link_inet6_get_addr_gen_mode(struct rtnl_link *,
 extern int		rtnl_link_inet6_set_addr_gen_mode(struct rtnl_link *,
 							  uint8_t);
 
+extern int		rtnl_link_inet6_get_flags(struct rtnl_link *,
+							  unsigned int *);
```
thom311: the implementation uses type `uint32_t`. It must be consistent, and `uint32_t` seems better as it's closer to the actual value. Same for the setter.

libnl-route-3.sym
matra01: I moved the functions to the new section libnl_3_4 as successor for section libnl_3_2_28. Is this correct?

include/netlink/route/link/inet6.h
matra01: I have changed parameter types to uint32_t [*] to meet the implementation.

libnl-route-3.sym
thom311: `rtnl_link_inet6_flags2str` and `rtnl_link_inet6_str2flags` as well.

matra01: Yes, I have seen it, as I hit the comment button. Fixed.

thom311:
I wonder if `rtnl_link_inet6_str2flags()` and flags2str also should have types `uint32_t` instead of `int`. Dunno. Maybe yes. But I am fine with the current state of the branch (unless you want to change this).

Will merge after 3.3.0 is out.

thom311:
merged as https://github.com/thom311/libnl/commit/b9a5a832330fd8cb28bddcc84840f445ad5da3c4

(actually, I messed up at first and pushed something wrong: https://github.com/thom311/libnl/commit/e0dc991ad5abe53e5ad7ba3762c1c993920c6641 )

PR #138, opened 2017-03-20
Provide accessors for actions (rtnl_act).

shanmugh:
Add accessor functions to retrieve actions on a tc object
and provide additional function to navigate the collection
of actions.

### DIFF
--- a/lib/route/cls/basic.c
+++ b/lib/route/cls/basic.c
@@ -233,6 +233,21 @@ int rtnl_basic_add_action(struct rtnl_cls *cls, struct rtnl_act *act)
 	return rtnl_act_append(&b->b_act, act);
 }
 
+struct rtnl_act* rtnl_basic_get_action(struct rtnl_cls *cls)
+{
+	struct rtnl_basic *b;
+
+	if (!(b = rtnl_tc_data_peek(TC_CAST(cls))))
+		return NULL;
+
+	if (!(b->b_mask & BASIC_ATTR_ACTION))
+		return NULL;
+
+	struct rtnl_act* act = b->b_act;
+    rtnl_act_get(act);
+    return act;
+}
+
 int rtnl_basic_del_action(struct rtnl_cls *cls, struct rtnl_act *act)
 {
 	struct rtnl_basic *b;
--- a/lib/route/cls/basic.c
+++ b/lib/route/cls/basic.c
@@ -243,9 +243,7 @@ struct rtnl_act* rtnl_basic_get_action(struct rtnl_cls *cls)
 	if (!(b->b_mask & BASIC_ATTR_ACTION))
 		return NULL;
 
-	struct rtnl_act* act = b->b_act;
-    rtnl_act_get(act);
-    return act;
+	return b->b_act;
 }
 
 int rtnl_basic_del_action(struct rtnl_cls *cls, struct rtnl_act *act)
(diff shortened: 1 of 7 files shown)


lib/route/cls/basic.c
```diff
@@ -233,6 +233,21 @@ int rtnl_basic_add_action(struct rtnl_cls *cls, struct rtnl_act *act)
...
+{
+	struct rtnl_basic *b;
+
+	if (!(b = rtnl_tc_data_peek(TC_CAST(cls))))
+		return NULL;
+
+	if (!(b->b_mask & BASIC_ATTR_ACTION))
+		return NULL;
+
+	struct rtnl_act* act = b->b_act;
```
congwang: In C, it is preferred to to move all declarations to the beginning of the function.

lib/route/cls/basic.c
```diff
@@ -233,6 +233,21 @@ int rtnl_basic_add_action(struct rtnl_cls *cls, struct rtnl_act *act)
...
+    rtnl_act_get(act);
```
congwang: Fix indention

lib/route/cls/u32.c
```diff
@@ -560,6 +560,21 @@ int rtnl_u32_add_action(struct rtnl_cls *cls, struct rtnl_act *act)
...
+    struct rtnl_u32 *u;
+
+    if (!(u = rtnl_tc_data_peek(TC_CAST(cls))))
+        return NULL;
+
+    if (!(u->cu_mask & U32_ATTR_ACTION))
+        return NULL;
+
+    struct rtnl_act* act = u->cu_act;
+    rtnl_act_get(act);
```
congwang: This grabs a refcount, so you need to provide an API to release that refcount too?

libnl-route-3.sym
```diff
@@ -822,6 +824,7 @@ global:
 	rtnl_u32_add_key_uint16;
 	rtnl_u32_add_key_uint32;
 	rtnl_u32_add_key_uint8;
+	rtnl_u32_get_action;
```
thom311: let's insert new lines sorted alphabetically.

lib/route/cls/u32.c
thom311: That API exists as `rtnl_act_put()`. But I don't think that `get_action()` should increase the ref-count at all. Just return the reference without incrementing the ref-count.

shanmugh: We already have `rtnl_put_act`.

lib/route/cls/basic.c
shanmugh: Fixed.

lib/route/cls/basic.c
shanmugh: Done.

libnl-route-3.sym
shanmugh: Fixed.

lib/route/cls/u32.c
congwang: @thom311 I mean if refcount is really needed rtnl_u32_put_action() is needed too, not that I miss rtnl_put_act()... But yeah, it looks like refcount is not necessary here.

shanmugh:
The new API is needed to fix an issue in Mesos where we need to query the actions on a filter. See https://issues.apache.org/jira/browse/MESOS-7266.

thom311:
Sorry for taking so long. I was rather busy these weeks.

I merged it now to master, but it missed 3.3.0 (which my fault). I didn't want to postpone 3.3.0 any longer and I also disliked the idea of adding new API after 3.3.0-rc1.

Let's see to do a 3.4.0 release earlier then planned.

PR #141, opened 2017-05-04
Revert "build: enable building cli during tests"

baruchsiach:
This reverts commit 3cb28534d34392ceec4adead0cfa97039796ccb7.

Contrary to what 3cb28534d commit log claims, the cli programs depend on
dynamic libraries support of the toolchain. Enabling cli programs
unconditionally breaks static build as follows:

In file included from lib/cli/cls/basic.c:12:0:
./include/netlink/cli/utils.h:25:19: fatal error: dlfcn.h: No such file or directory
compilation terminated.
Makefile:3666: recipe for target 'lib/cli/cls/lib_cli_cls_basic_la-basic.lo' failed
make[1]: *** [lib/cli/cls/lib_cli_cls_basic_la-basic.lo] Error 1

Revert that commit to restore the ability of static only build of libnl, and
its dependencies.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -3,8 +3,6 @@
 ACLOCAL_AMFLAGS = -I m4
 
 lib_LTLIBRARIES =
-noinst_LTLIBRARIES =
-check_LTLIBRARIES =
 
 check_PROGRAMS =
 check_programs =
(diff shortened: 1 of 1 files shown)


thom311:
I'd rather not revert the patch. It's there so that during `make check` we build the code (that we can build).

If you want to bulid cli without dlfcn.h, I think the correct solution is to allow for that.

How about instead https://github.com/thom311/libnl/commits/cli-no-dlfcn-h ?

thom311:
merged commit https://github.com/thom311/libnl/commit/dad421eaac61cde08945a4fc42f4cd16d2ad5607 .

I think that solves this issue. Please reopen, if there is a remaining issue. Thanks

baruchsiach:
Commits 144c6c5e6a31 and c881a27fbda fix static build for me.

Thanks.

PR #144, opened 2017-05-23
genl: drop usage of GENL_ID_GENERATE

tklauser:
After kernel commit a07ea4d9941a ("genetlink: no longer support using
static family IDs"), GENL_ID_GENERATE is no longer exposed to userspace
(and actually should never have been). Update the private header copy of
linux/genetlink.h accordingly. And replace the two occurences of
GENL_ID_GENERATE.


thom311:
merged as 32d13c0058fd1beacb091c1d57999854819c3bc2

Thanks Tobias!!

PR #145, opened 2017-06-08
realloc fixes

Villemoes:
These fix a few standard realloc memleak bugs, with a few cleanups thrown in that I noticed while skimming the code (and since I'm not familiar with the code base, I'm not 100% certain they're all valid, but they seem fairly obvious).

### DIFF
--- a/lib/cache_mngr.c
+++ b/lib/cache_mngr.c
@@ -317,15 +317,18 @@ retry:
 			break;
 
 	if (i >= mngr->cm_nassocs) {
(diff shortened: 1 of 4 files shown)


thom311:
Nice work!

I added 3 more patches on top of your branch. Please check if they are OK.

Merged as https://github.com/thom311/libnl/commit/429efd4ebaeebf32ad594f79a3e125ab5257a458

Villemoes:
Seems ok at a quick glance.

PR #147, opened 2017-07-01
genl: fix potential NULL pointer in genl_unregister_family()

up2wing:
Though this may hardly occur, but it may be better to do this
to avoid application segfault?

Signed-off-by: Yi Wang <wang.yi59@zte.com.cn>


up2wing:
eh~ my fault,  I mistook the o_list for a pointer...
sorry for the trouble :(

PR #151, opened 2017-09-07
cli: include sys/select.h for select(2)

tklauser:
Some of the cli tools use select(2) and its man page states:

  /* According to POSIX.1-2001, POSIX.1-2008 */
  #include <sys/select.h>

Do so and explicitly #include <sys/select.h> in <netlink/cli/utils.h>
instead of relying of getting select(2) via implicit includes. This is
also needed to make libnl compile for Android.

Based on a previous patch by Fredrik Fornwall.

### DIFF
--- a/include/netlink/cli/utils.h
+++ b/include/netlink/cli/utils.h
@@ -24,6 +24,7 @@
 #include <getopt.h>
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/select.h>
 
 #include <netlink/netlink.h>
 #include <netlink/utils.h>


thom311:
Hi Tobias,

merged as https://github.com/thom311/libnl/commit/4e957ea615456c5f076707f56c422af2b70f0188

`include/netlink/cli/utils.h` is a public header file. In principle I would prefer to only include headers that are required to make the headers self-contained. I opted to merge your patch nonetheless because:

  - `utils.h` already includes  *a lot* or other headers, so there are prior examples.

  - cli devel packages are probably not used much outside of building libnl3 cli itself. I think it has few users, so the let's relax the standard here a bit. 

  - it's just less work this way, for something which is probably cosmetical only.

Thanks!!

PR #155, opened 2017-10-16
lib/route/link/vlan.c: fix memory corruption in rtnl_link_vlan_set_egress_map

sbixl:
Hi,

if you set more than four entries for the vlan egress map a memory corruption occurs because
the reallocation does not reserve memory for more than four.

In my use case I've to configure several interfaces where all possible pcp values in the  vlan tag must be mapped in the right way, so eight entries must be set using rtnl_link_vlan_set_egress_map. The issue is reproducible on several platforms (ARM, x86_64, ...)

Fixing the issue by using vi_negress instead of vi_egress_size used by calloc and realloc and removing the magic value "32" which I didn't understand so far what was the background/intention of it. Maybe someone can tell me? :-)

regards Sebastian

### DIFF
--- a/lib/route/link/vlan.c
+++ b/lib/route/link/vlan.c
@@ -162,11 +162,11 @@ static int vlan_parse(struct rtnl_link *link, struct nlattr *data,
 			i++;
 		}
 
-		/* align to have a little reserve */
-		vi->vi_egress_size = (i + 32) & ~31;
-		vi->vi_egress_qos = calloc(vi->vi_egress_size, sizeof(*vi->vi_egress_qos));
+		vi->vi_egress_qos = calloc(i, sizeof(struct vlan_map));
 		if (vi->vi_egress_qos == NULL)
 			return -NLE_NOMEM;
+
+		vi->vi_egress_size = i * sizeof(struct vlan_map);
 
 		i = 0;
 		nla_for_each_nested(nla, tb[IFLA_VLAN_EGRESS_QOS], remaining) {
@@ -270,13 +270,13 @@ static int vlan_clone(struct rtnl_link *dst, struct rtnl_link *src)
 		return err;
 	vdst = dst->l_info;
 
-	vdst->vi_egress_qos = calloc(vsrc->vi_egress_size,
+	vdst->vi_egress_qos = calloc(vsrc->vi_negress,
 				     sizeof(struct vlan_map));
 	if (!vdst->vi_egress_qos)
 		return -NLE_NOMEM;
 
 	memcpy(vdst->vi_egress_qos, vsrc->vi_egress_qos,
-	       vsrc->vi_egress_size * sizeof(struct vlan_map));
+	       vsrc->vi_negress * sizeof(struct vlan_map));
 
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


thom311:
The 32 is just a constant growth value. Maybe a factor like time 1.5 would be better. Anyway, it's not wrong.

I think, the code becomes more complicated by changing the meaning of `vi_egress_size` to contain the number of bytes, instead of the number of elements.

A simple fix would be

```
-  ptr = realloc(vi->vi_egress_qos, new_size);
+  ptr = realloc(vi->vi_egress_qos, new_size * sizeof(struct vlan_map));
```

I think `vlan_clone()` is also wrong, as it doesn't copy anything `vi_egress_qos`.

sbixl:
Hi Thomas,

I think there was a misunderstanding by interpreting `vi_negress` and `vi->vi_egress_size` from my side, sorry for that.

After debugging the code the first time, I became struggled by the condition `if (vi->vi_negress >= vi->vi_egress_size)`. The parameter `vi->vi_negress` counts the elements currently stored in the map and the parameter `vi->vi_egress_size` is updated by addition with 32. The first time adding a entry the condition is true and a reallocation takes place where 32 Bytes are reserved. Now we are able to store
four entries (because one entry needs eight Byte) until the next reallocation shall take place. 
But before adding the fourth entry the condition doesn't match because `vi->vi_negress` is 4 and vi`->vi_egress_size` is 32. At this point I assumed there is something wrong with parameter `vi->vi_egress_size` and I don't thought `vi->vi_egress_size` was related to the max. elements the map can store because the vlan pcp can only be in a range from 0 to 7 (like specified in IEEE 802.1Q) and therefore only eight mappings are reasonable.

After reading your comment it's clear now that `vi->vi_egress_size` is used to provide the information about the actual maximum amount of entries which can be stored in the current map.

After adding your proposal and testing it, the crash doesn't happen anymore, but now we will reallocate a map which can hold up to 32 entries but at least a maximum of eight shall be possible. You can simply check this by a short test reading back the egress map and you will get back every time eight entries even if you set more than eight previously.

Should it not even better to change the 32 for instance to four so a new reallocation will take place if the user wants to add more than four entries or should we work with a fixed array like it's done for the ingress map. In addition there should be a check if more than eighth entries becomes added by the user?

Regarding vlan_clone() you alright this should be fixed too.

Do you agree with my suggestion if yes I will prepare an new solution for review.

regards 
sebastian

thom311:
> After reading your comment it's clear now that vi->vi_egress_size is used to provide the information 
> about the actual maximum amount of entries which can be stored in the current map.

`vi_egress_size` is the number of allocated entires, while `vi_negress` is the number of actually used entires. Neither are in bytes, both count the "struct vlan_map" entries.

>  but now we will reallocate a
> map which can hold up to 32 entries but at least a maximum of eight shall be possible.

I think for the egress map, you can add significantly more then 8 entries. Only ingress map is limited to 8 entries.

So, when you add entries one by one, you need to increase the buffer size. Usually that is done by growing the size exponentially (e.g. doubling each time, or times 1.5). In this case, it grows by a constant amount each time of 32 entries (not bytes).
Growing by a constant number is generally bad, it would be better to do:

    uint32_t new_size = vi->vi_egress_size + 1 + vi->vi_egress_size / 2;


Let's first do one simple patch that fixes the memory corruption (the realloc line). Then if you would like, change how to calculate new_size and fix clone() (possibly in separate patches). Does that make sense?

sbixl:
I agree, first the memory corruption should be fixed by a simple patch you mentioned. I will do some further investigation to figure out if is really true to add significantly more then 8 entries to the egress map because this doesn't make really sense to me (from logical point of view). Thank you for the great conversation. :-)

thom311:
Merged the combined patch to master + 3 follow-up patches ( https://github.com/thom311/libnl/commit/56229168614a11897bd620c363e71aca3faf23ee )

Closing PR as fixed. Testing is appreciated.

Thanks Sebastian

PR #156, opened 2017-11-02
Enable setting output interface attribute on route objects

sargun:
Previously, you could not directly set RTA_OIF from the public
route object modification API, this reveals that attribute.

Signed-off-by: Sargun Dhillon <sargun@sargun.me>

### DIFF
--- a/include/netlink-private/types.h
+++ b/include/netlink-private/types.h
@@ -341,6 +341,7 @@ struct rtnl_route
 	struct nl_addr *	rt_src;
 	uint32_t		rt_table;
 	uint32_t		rt_iif;
+	uint32_t		rt_oif;
 	uint32_t		rt_prio;
 	uint32_t		rt_metrics[RTAX_MAX];
 	uint32_t		rt_metrics_mask;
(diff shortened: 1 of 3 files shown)


thom311:
Hm. RTA_OIF is part of the next-hop.

Usually, kernel would report `RTA_MULTIPATH`, in which case, you might have multiple next-hops. The first of these hops, is the one whose `rtnl_route_nh_set_ifindex()` equals `RTA_OIF`.

Why do you want to set this separately of the next-hops? Can you explain a bit better what this patch is trying to solve?

sargun:
If you have an unnumbered interface, you need this. For example, if you're running the iproute2 command:

```
ip link add type veth
ip link set 
ip route add 1.2.3.4/32 dev veth0
```

In that, it doesn't appear you pass a next-hop with the route object, but just an OIF for the route itself.

thom311:
> ip route add 1.2.3.4/32 dev veth0

in libnl3 you can express such a route, with a route that has one next-hop. With the next-hop having no gateway, but the ifindex set.

that should work, no?

PR #158, opened 2017-12-01
Fix bug in test-create-bridge.c

grahn:
The call to rtnl_link_get_master() at the end of the example can
misleadingly fail because the nl_cache isn't refilled after adding the
test interface to the test bridge.

This commit changes the example to refill the cache before calling
rtnl_link_get_master().


thom311:
reloading the entire cache after a change isn't very efficient or the way to do it in production code.

For these test programs, that's acceptable.

Merged, thanks!!

PR #159, opened 2017-12-20
route: add separate function to set netem qdisc delay distribution

stv0g:
A new function rtnl_netem_set_delay_distribution_data() has been added
to allow the user to pass the delay distribution directly without loading
it from a file.

In conjunction with the maketable code (see iproute2 / NISTnet) this can
be used to generate and load custom delay distributions on the fly.


thom311:
Hi Steffen,

thanks for the patch. Merged to master with a few follow up adjustments.

PR #161, opened 2018-01-03
tests: Add test to {de}activate loopback interface

marcosps:
This tests is much more like an example of how to do it, and also works
as a test to check if rtnl_link_change is working as expected when it
comes to loopback interface.

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -813,7 +813,8 @@ check_PROGRAMS += \
 	tests/test-create-vxlan \
 	tests/test-delete-link \
 	tests/test-socket-creation \
-	tests/test-u32-filter-with-actions
+	tests/test-u32-filter-with-actions \
+	tests/test-loopback-up-down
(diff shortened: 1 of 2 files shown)


marcosps:
I could not find where to wire up this new test... it seems that only check-all.c is being executed... am I missing something?

thom311:
rebased and merged.

Thanks!!

thom311:
> I could not find where to wire up this new test... it seems that only check-all.c is being executed... am I
> missing something?

Well, these ~tests~ are not executed as part of `make check`. They are more for manual testing and just whether it compiles.

Of course, that should be all better and there should be plenty of unit tests. However, as it is, that would at least require that the test runner creates a new network name space, so that the test has the required permissions/capabilities.

PR #162, opened 2018-01-04
lib/veth.c: Disassociate link name of peer name

marcosps:
Based in the function documentation, peer name could be set regardless
of the link name being informed of not. These parameters should be
checked independently.

Signed-off-by: Marcos Paulo de Souza <marcos.souza.org@gmail.com>


thom311:
Makes sense.

rebased and merged to master.

Thanks for the patch!!

PR #166, opened 2018-01-15
Add new function for setting parent of classifier cache.

d0u9:
It is not good to give classifier cache users only one chance to set parent id when allocte new cache. Sometimes we want to reuse classifier cache to reduce the overhead of allocating new memory everytime a new cache is created.

### DIFF
--- a/include/netlink/route/classifier.h
+++ b/include/netlink/route/classifier.h
@@ -27,6 +27,8 @@ extern void		rtnl_cls_put(struct rtnl_cls *);
 extern int		rtnl_cls_alloc_cache(struct nl_sock *, int, uint32_t,
 					     struct nl_cache **);
 
+extern void 		rtnl_cls_cache_set_tcm_params(struct nl_cache *, int, uint32_t);
+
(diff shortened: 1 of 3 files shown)


thom311:
New functions must be exported, by adding an entry to [libnl-route-3.sym](https://github.com/thom311/libnl/blob/2b11ab02ce24476f96c22b18e27f3737559010a2/libnl-route-3.sym#L1065).

Also, I think if you add such API, it should both set ifindex and parent at once. Meaning, rename the function (maybe `rtnl_cls_cache_set_tcm_params()`?), and add an `ifindex` argument as well.

d0u9:
I will make some improvements and re-commit later.

d0u9:
Patch is updated.

thom311:
> Patch is updated.

I think it is not updated, it's still 45644036897285aeb5617a0ae2fe97a7d8fa22d3. You need to `git push --force` to the branch to update it.

d0u9:
It should be OK now. I previously pushed changes to the wrong branch.

thom311:
rebased and merged to master, with a follow-up fix.

Thank you

PR #167, opened 2018-01-23
Dirty fix for libnl cgroup filter addition problem.

d0u9:
Currently, due to the incomplete netlink datagram sent by libnl, cgroup
filter addition is not fully functional. The datagram generated by `tc`
command includes an empty attribute section, which is stripped off
in the libnl counterpart.

In this dirty fix, we selectively disable the deletion for `cgroup`
filter.

### DIFF
--- a/include/netlink/attr.h
+++ b/include/netlink/attr.h
@@ -143,6 +143,7 @@ extern int		nla_put_nested(struct nl_msg *, int,
 				       const struct nl_msg *);
 extern struct nlattr *	nla_nest_start(struct nl_msg *, int);
 extern int		nla_nest_end(struct nl_msg *, struct nlattr *);
+extern int		nla_nest_end_keep_empty(struct nl_msg *, struct nlattr *);
 extern void		nla_nest_cancel(struct nl_msg *, const struct nlattr *);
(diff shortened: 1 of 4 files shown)


d0u9:
This maybe not a perfect solution for adding cgroup filter in the traffic control subsystem, but it does work.

thom311:
We cannot just omit calling `nla_nest_end()`. If we require that `nla_nest_end()` preserves an empty attribute, we need a new function (maybe `nla_nest_end_keep_empty()`). Or something like that.

d0u9:
`nla_nest_end_keep_empty()`? I think we need `rtnl_tc_msg_build_keep_empty()`. However, it is hard to branch the calling path when we need to handle some **strange** objects. It is ugly to simply insert `if-else` statements here.

thom311:
Why do you think so? Why can you not do:

```
	if (strcmp("cgroup", tc->tc_kind))
		nla_nest_end(msg, opts);
	else
		nla_nest_end_keep_empty(msg, opts);
```

d0u9:
Well, it looks ok. I'll push another patch later.

PR #168, opened 2018-02-11
list.h: simplify element access macro, remove use of non-portable GCC compound statement

tkoeppe:


thom311:
The current version of the macro enforces that the pointer types match via
```
const typeof( ((type *)0)->member ) *__mptr = (ptr);
```
That is a desired compile time check.

Both clang and gcc support typeof(), and we don't test any other compilers. Does this solve an actual problem or do you just dislike non standard C extensions?

tkoeppe:
I see. No, I just thought of it as a simplification. If you want to preserve the check, then indeed the status quo is preferable. Never mind then!

PR #169, opened 2018-02-14
route: Fix IPv6 default prio

scaidermern:
rtnl_route_parse assigned the wrong default metric to IPv6 routes for
netlink messages not containing a route attribute. This lead to some
IPv6 routes being reported twice since the prio is part of the route
cache key.


scaidermern:
I just noticed that this fix isn't enough. Some IPv6 routes are still reported with a wrong prio (0 instead of 1024).

scaidermern:
I was wrong. I looked at cloned routes (flag RTM_F_CLONED aka 512) but now I'm pretty sure that I just want to ignore them :)

PR #170, opened 2018-02-22
matchall classifier

voldymyr:
This patch set adds "matchall" classifier support for libnl. More information about matchall classifier is here: http://man7.org/linux/man-pages/man8/tc-matchall.8.html
What are your thoughts/comments? Let me know if you would like anything in the code to be changed.

Thanks!


lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+	mall->m_mask |= MALL_ATTR_CLASSID;
+
+	return 0;
+}
+
+int rtnl_mall_get_classid(struct rtnl_cls *cls, uint32_t *classid)
+{
+	struct rtnl_mall *mall;
+
+	if (!(mall = rtnl_tc_data(TC_CAST(cls))))
```
thom311: could we use `rtnl_tc_data_peek` to avoid allocating the memory? Then, also the failure case is `-NLE_EINVAL` instead of `-NLE_NOMEM`.

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+}
+
+int rtnl_mall_set_flags(struct rtnl_cls *cls, uint32_t flags)
+{
+	struct rtnl_mall *mall;
+
+	if (!(mall = rtnl_tc_data(TC_CAST(cls))))
+		return -NLE_NOMEM;
+
+	mall->m_flags |= flags;
```
thom311: there is no API to clear flags (like `rtnl_mall_unset_flags`). But I don't think we need that anyway, do we? Just set them all and let the caller figure out which flags to specify
```
mall->m_flags = flags;
```

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+	mall->m_mask |= MALL_ATTR_FLAGS;
+
+	return 0;
+}
+
+int rtnl_mall_get_flags(struct rtnl_cls *cls, uint32_t *flags)
+{
+	struct rtnl_mall *mall;
+
+	if (!(mall = rtnl_tc_data(TC_CAST(cls))))
```
thom311: peek

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+		return -NLE_NOMEM;
+
+	if (!(mall->m_mask & MALL_ATTR_FLAGS))
+		return -NLE_INVAL;
+
+	*flags = mall->m_flags;
+	return 0;
+}
+
+int rtnl_mall_add_action(struct rtnl_cls *cls, struct rtnl_act *act)
```
thom311: this really appends an action and one can add multiple actions. Shouldn't it be called `rtnl_mall_append_action()`?

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+
+	if (!act)
+		return 0;
+
+	if (!(mall = rtnl_tc_data(TC_CAST(cls))))
+		return -NLE_NOMEM;
+
+	mall->m_mask |= MALL_ATTR_ACTION;
+	rtnl_act_get(act);
+	return rtnl_act_append(&mall->m_act, act);
```
thom311: If `rtnl_act_append` fails, you must release the reference. Or maybe better: only `rtnl_act_get(act)` after `rtnl_act_append` succeeded.

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+}
+
+struct rtnl_act *rtnl_mall_get_action(struct rtnl_cls *cls)
```
thom311: maybe `rtnl_mall_get_first_action()`?

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+		return 0;
+
+	if (!(mall = rtnl_tc_data(TC_CAST(cls))))
+		return -NLE_NOMEM;
+
+	if (!(mall->m_mask & MALL_ATTR_ACTION))
+		return -NLE_INVAL;
+
+	ret = rtnl_act_remove(&mall->m_act, act);
+	if (ret)
```
thom311: Should be `if (ret < 0)` (although, in practice, it makes no difference, because rtnl_act_remove() never returns a positive number for the moment).

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	if (tb[TCA_MATCHALL_FLAGS]) {
+		mall->m_flags = nla_get_u32(tb[TCA_MATCHALL_FLAGS]);
+		mall->m_mask |= MALL_ATTR_FLAGS;
+	}
+
+	if (tb[TCA_MATCHALL_ACT]) {
+		mall->m_mask |= MALL_ATTR_ACTION;
+		err = rtnl_act_parse(&mall->m_act, tb[TCA_MATCHALL_ACT]);
```
thom311: here, if the netlink message contains a `tb[TCA_MATCHALL_ACT]` attribute, but zero actions, then we will set `mall->m_mask |= MALL_ATTR_ACTION;`. Compare this to `rtnl_mall_del_action`, which clears `MALL_ATTR_ACTION` when removing the last action. It should be consistent: either, having no action means that the attribute flag is cleared, or once we set the attribute, we never clear it (meaning: that also an explicit empty list of actions, counts like actions are set). Probably, the second notion makes more sense, because in `mall_msg_fill` it determines whether to write a `TCA_MATCHALL_ACT` attribute.

lib/route/cls/mall.c
```diff
@@ -0,0 +1,288 @@
...
+	struct rtnl_mall *dst = _dst, *src = _src;
+
+	if (src->m_act) {
+		if (!(dst->m_act = rtnl_act_alloc()))
+			return -NLE_NOMEM;
+
+		memcpy(dst->m_act, src->m_act, sizeof(struct rtnl_act));
+
+		/* action nl list next and prev pointers must be updated */
+		nl_init_list_head(&dst->m_act->ce_list);
```
thom311: this only copies the first action. Is that intentional? It should deep-clone the entire list.

include/linux-private/linux/pkt_cls.h
```diff
@@ -464,4 +464,16 @@ enum {
 	TCF_EM_OPND_LT
 };
 
+/* Match-all classifier */
+
```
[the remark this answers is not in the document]
voldymyr: Ok, I will copy the entire header.

lib/route/cls/mall.c
voldymyr: Sure, thx for advice.

lib/route/cls/mall.c
voldymyr: I agree, we can unset flags by passing '0' as parameter to *set_flags(). I'll fix this as you mentioned.

lib/route/cls/mall.c
voldymyr: You're right, 'append' sounds more reasonable in this case.

lib/route/cls/mall.c
voldymyr: Great, thx for spotting this.

lib/route/cls/mall.c
voldymyr: Sure, I will rename it.

lib/route/cls/mall.c
voldymyr: ok

voldymyr:
I made corrections according to your comments, please have a look whenever you have time. Thanks :)

thom311:
Hi, it looks mostly good to me now.

`mall_clone()` though still seems wrong to me. Shouldn't it be?

```diff
From deef9f29b000ac19d0412ad66cc3411fc19bd394 Mon Sep 17 00:00:00 2001
From: Thomas Haller <thaller@redhat.com>
Date: Fri, 6 Apr 2018 16:42:55 +0200
Subject: [PATCH 1/1] route/mall: fix deep cloning mall

rtnl_act_append() does not clone the object, it takes it over
and thus the source object is destroyed.
---
 lib/route/cls/mall.c | 16 ++++++++++------
 1 file changed, 10 insertions(+), 6 deletions(-)

diff --git a/lib/route/cls/mall.c b/lib/route/cls/mall.c
index d1588f65..e13ee92f 100644
--- a/lib/route/cls/mall.c
+++ b/lib/route/cls/mall.c
@@ -225,25 +225,29 @@ static int mall_msg_fill(struct rtnl_tc *tc, void *data, struct nl_msg *msg)
 static int mall_clone(void *_dst, void *_src)
 {
 	struct rtnl_mall *dst = _dst, *src = _src;
-	struct rtnl_act *next;
+	struct rtnl_act *next, *new;
 	int err;
 
 	if (src->m_act) {
 		if (!(dst->m_act = rtnl_act_alloc()))
 			return -NLE_NOMEM;
 
+		/* action nl list next and prev pointers must be updated */
+		nl_init_list_head(&dst->m_act->ce_list);
+
 		memcpy(dst->m_act, src->m_act, sizeof(struct rtnl_act));
 		next = rtnl_act_next(src->m_act);
 		while (next) {
-		        err = rtnl_act_append(&dst->m_act, next);
+			new = (struct rtnl_act *) nl_object_clone((struct nl_object *) next);
+			if (!new)
+				return -NLE_NOMEM;
+
+			err = rtnl_act_append(&dst->m_act, new);
 			if (err < 0)
-		                return err;
+				return err;
 
 			next = rtnl_act_next(next);
 		}
-
-		/* action nl list next and prev pointers must be updated */
-		nl_init_list_head(&dst->m_act->ce_list);
 	}
 
 	return 0;
-- 
2.14.3
```


Also, could you squash all the fixups, and force-push the branch? So, that there are just two remaining commits, like you did originally.

That'd be all :)

voldymyr:
Branch is now updated, commits squashed. Thx for fixing mall_clone() and generally for the review.

thom311:
Merged as https://github.com/thom311/libnl/commit/da4e707af8059fb08ac9233440732923984993dd

Thanks!!

PR #173, opened 2018-03-04
build: fix build on musl libc

karlguy:
This PR fixes portability problems such as dependencies on glibc macros. This patch has been tested against musl and glibc, and builds with no issues.

### DIFF
--- a/include/linux-private/linux/if_ether.h
+++ b/include/linux-private/linux/if_ether.h
@@ -22,6 +22,7 @@
 #define _LINUX_IF_ETHER_H
 
 #include <linux/types.h>
+#include <linux/libc-compat.h>
 
 /*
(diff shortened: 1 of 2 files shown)


thom311:
The files under `include/linux-private` are supposed to be direct copies of the user space kernel headers.

Of course, they diverge from the original version in kernel as time goes by, but they should be updated by re-importing newer versions without manual adjustments.

Where did you get header files? Did you edit them manually?

thom311:
Does musl work with latest user-space headers from upstream kernel? If it does, you can just re-import the latest headers. If it doesn't, it seems like a kernel/musl bug and should be fixed there.

karlguy:
Damn, it looks like the upstream kernel does not have this patch applied, so I will bring this patch there. Thanks!

PR #176, opened 2018-03-27
Add support for cloning cgroup filter object.

d0u9:
In this commit, we implement ematch_tree_clone(), which is basis of
cgroup_clone() interface. The whole ematch tree is deep-copied except
the e_ops filed.

Also, a new test is added for testing the interface, which named as
test-ematch-tree-clone.c located in tests directory.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -896,7 +896,8 @@ tests_check_all_SOURCES = \
 	tests/util.h \
 	tests/check-all.c \
 	tests/check-addr.c \
-	tests/check-attr.c
+	tests/check-attr.c \
+	tests/check-ematch-tree-clone.c
 
 tests_check_all_CPPFLAGS = \
 	$(tests_cppflags) \
--- a/lib/route/cls/ematch.c
+++ b/lib/route/cls/ematch.c
@@ -289,6 +289,63 @@ void rtnl_ematch_tree_free(struct rtnl_ematch_tree *tree)
 	free(tree);
 }
 
+static int clone_ematch_list(struct nl_list_head *dst, struct nl_list_head *src)
+{
+	struct rtnl_ematch *new = NULL, *pos = NULL;
+
+	nl_list_for_each_entry(pos, src, e_list) {
+		new = rtnl_ematch_alloc();
+		if (!new)
+			goto nomem;
+
+		new->e_id      = pos->e_id;
+		new->e_kind    = pos->e_kind;
+		new->e_flags   = pos->e_flags;
+		new->e_index   = pos->e_index;
+		new->e_datalen = pos->e_datalen;
+
+		if (pos->e_ops) {
+			if (rtnl_ematch_set_ops(new, pos->e_ops))
+				goto nomem;
+		}
+
+		if (!nl_list_empty(&pos->e_childs)) {
+			if (clone_ematch_list(&new->e_childs, &pos->e_childs) < 0)
+				goto nomem;
+		}
+		nl_list_add_tail(&new->e_list, dst);
+	}
+
+	return 0;
+
+nomem:
+	if (new)
+		free(new);
+	free_ematch_list(dst);
+	return -NLE_NOMEM;
+}
+
+/**
+ * Clone ematch tree object
+ * @arg src		ematch tree object
+ *
+ * This function clones the ematch tree and all ematches attached to it.
+ */
+struct rtnl_ematch_tree *rtnl_ematch_tree_clone(struct rtnl_ematch_tree *src)
+{
+	struct rtnl_ematch_tree *dst = NULL;
+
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -817,6 +817,7 @@ check_PROGRAMS += \
 	tests/test-delete-link \
 	tests/test-socket-creation \
 	tests/test-u32-filter-with-actions \
+	tests/test-ematch-tree-clone \
```
thom311: please sort the new name alphabetically (yes, it's not entirely sorted yet, but new entries should be sorted. Optionally, a follow-up patch to sort it all would be nice).

Makefile.am
```diff
@@ -863,6 +864,8 @@ tests_test_complex_HTB_with_hash_filters_CPPFLAGS = $(tests_cppflags)
 tests_test_complex_HTB_with_hash_filters_LDADD    = $(tests_ldadd)
 tests_test_u32_filter_with_actions_CPPFLAGS       = $(tests_cppflags)
 tests_test_u32_filter_with_actions_LDADD          = $(tests_ldadd)
+tests_test_ematch_tree_clone_CPPFLAGS             = $(tests_cppflags)
+tests_test_ematch_tree_clone_LDADD                = $(tests_ldadd)
```
thom311: also sort into the existing lines.

libnl-route-3.sym
```diff
@@ -204,6 +204,7 @@ global:
 	rtnl_ematch_tree_alloc;
 	rtnl_ematch_tree_dump;
 	rtnl_ematch_tree_free;
+	rtnl_ematch_tree_clone;
```
thom311: The already release version cannot be changed by adding new symbols to it. New symbols must always be added to the version for the next-coming release (`libnl_3_5` at the moment), at the bottom of the file.

tests/test-ematch-tree-clone.c
```diff
@@ -0,0 +1,137 @@
...
+	struct rtnl_ematch *child = NULL;
+
+	if (!parent)
+		return 0;
+
+	if (++current_depth > MAX_DEPTH) {
+		--current_depth;
+		return 0;
+	}
+
```
thom311: trailing whitespace

tests/test-ematch-tree-clone.c
```diff
@@ -0,0 +1,137 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
```
thom311: Is this necessary? Why?

lib/route/cls/ematch.c
```diff
@@ -289,6 +289,65 @@ void rtnl_ematch_tree_free(struct rtnl_ematch_tree *tree)
...
+
+	if (!src)
+		return NULL;
+
+	if (!(dst = rtnl_ematch_tree_alloc(src->et_progid)))
+		return NULL;
+
+	clone_ematch_list(&dst->et_list, &src->et_list);
+
+	NL_DBG(2, "Cloned ematch tree %p\n", src);
```
thom311: It this kind of logging really useful? Maybe not.

lib/route/cls/ematch.c
```diff
@@ -289,6 +289,65 @@ void rtnl_ematch_tree_free(struct rtnl_ematch_tree *tree)
...
+		new->e_index   = pos->e_index;
+		new->e_datalen = pos->e_datalen;
+
+		if (pos->e_ops) {
+			if (rtnl_ematch_set_ops(new, pos->e_ops))
+				goto nomem;
+		}
+
+		if (!nl_list_empty(&pos->e_childs)) {
+			if (clone_ematch_list(&new->e_childs, &pos->e_childs))
```
thom311: Check for failure with `< 0`.

thom311: Because, errors should be signaled by negative integers and the preferred style for checking whether an integer return value signals an error is `if (r < 0)`. Arguably, if you look at `clone_ematch_list` it never returns a positive value. But one shouldn't need to look at `clone_ematch_list` to verify that.

thom311: ah, if you find places where this is not honoured, that is a bug. Likely it's not a serious bug, only violating the style. Those places should be fixed, but well. Let's get it at least that way for new code.

Makefile.am
d0u9: Done

d0u9:
Patch updated.

d0u9:
Any update?

thom311:
sorry for the very late reply.

Yes, patch looks good to me. Very nice, especially thank you for adding a test!!

merged: ddfc36f46e2d1581e2b08bbe167644bd8b1f1571

PR #180, opened 2018-04-05
neigh: set correct AF for NDA_DST

toanju:
In case using a VXLAN interface at a bridge you will set L2 bridging
entries using a IP destination to tunnel the according L2 traffic. The
current behavior for the dst entries for a neighbor is to use the AF of
the neighbor itself thus in this case AF_BRIDGE is set. This is changed
in the PR to update the family of the dst using nl_addr_guess_family.


thom311:
Merged. Thank you!!

PR #182, opened 2018-04-06
neigh: support bridge entries for vxlan interfaces

toanju:
bridge entries used for switching into vxlan interfaces do not include a
vlan. A comparison of such entires currently always fails which leads
to an invalid cache. This patch selectively adds the NEIGH_ATTR_VLAN
flag based on the passed entry.


thom311:
Merged.

Thank you!!

PR #184, opened 2018-04-09
route/link: Fix #179

### LINKED ISSUE #179: rtnl_link_change returns NLE_SEQ_MISMATCH
Context: I am trying to set the MAC address of an interface using `rtnl_link_change`.  Here is the relevant snippet:

```cpp
int Foo::bar(struct rtnl_link *link, struct nl_addr *laddr) {
    struct rtnl_link *newlink = rtnl_link_alloc();
    if (newlink == NULL)
        return NLE_NOMEM;
    rtnl_link_set_addr(newlink, laddr);
    return rtnl_link_change(this->nl_sock, link, newlink, NLM_F_CREATE | NLM_F_REPLACE);
}
```

I am consistently getting `NLE_SEQ_MISMATCH` out of this, but it is successful, as my interface's MAC is definitely changed. `rtnl_link_change` is documented to resend the message as a `RTM_SETLINK` for compatibility reasons if `RTM_NEWLINK` yields a `NLE_OPNOTSUPP`. Indeed, the following simpler code always returns a not supported error:

```cpp
int Foo::bar(struct rtnl_link *link, struct nl_addr *laddr) {
    rtnl_link_set_addr(link, laddr);
    return rtnl_link_add(this->nl_sock, link, NLM_F_CREATE | NLM_F_REPLACE);
}
```

From the implementation of `rtnl_link_change`, there is indeed a discrepancy in sequence numbers when retrying to send the message. Each attempt uses `nl_send_auto`, which increments the sequence number of the socket. However, the sequence number of the message itself is *not* incremented, which leads to the incoherency. It was set by the first call to `nl_send_auto` which calls `nl_complete_msg`, but it will not set it again on second call since it has already been set.

The easiest way would be resetting `nlmsg_seq` to `NL_AUTO_SEQ` in the message headers when setting the message type to `RTM_SETLINK`.

I am willing to submit a PR, and will ignore the error for now u_u

Tuetuopay:
Handle properly sequence numbers in `rtnl_link_change`

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -1772,6 +1772,10 @@ errout:
  * @note The link name can only be changed if the link has been put
  *       in opertional down state. (~IF_UP)
  *
+ * @note On versions up to 3.4.0, \c NLE_SEQ_MISMATCH would be returned if the
+ *       kernel does not supports \c RTM_NEWLINK. It is advised to ignore the
+ *       error code if you cannot upgrade the library.
+ *
  * @return 0 on success or a negative error code.
  */
 int rtnl_link_change(struct nl_sock *sk, struct rtnl_link *orig,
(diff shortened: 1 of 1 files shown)


thom311:
merged with two minor follow-up fixes as https://github.com/thom311/libnl/commits/2a3a66977936471b93526adda7faa1ac231f78d7

Thanks!!

PR #185, opened 2018-04-10
Add new api rtnl_class_get_by_parent()

d0u9:
This function searches a class cache previously allocated with rtnl_class_alloc_cache()
and searches for a class matching the interface index and parent qdisc.


thom311:
merged as https://github.com/thom311/libnl/commit/17f0459c8d1891473c315315c0f5c25f38b24d6b

Thanks!!

PR #187, opened 2018-04-17
neigh: correct symbol exposed

toanju:
instead of exposing rtnl_neigh_ll_get the function
rtnl_neigh_get_by_vlan has to be exposed

fixes 3bf503d3


thom311:
merged as 0a96d14417ecdba8218d2c967e3c543200130bc6

thanks Tobias!!

PR #190, opened 2018-04-20
neigh_dump_line: dump master as well

toanju:
In case NEIGH_ATTR_MASTER is set this dumps as well the master
interface.


thom311:
merged as 3ebb0ea66b165d061c287164ca4514c13f8b2772

PR #192, opened 2018-05-15
route/link/vxlan: Fix IPv4 set_local resetting ce_mask

Tuetuopay:
`rtnl_link_vxlan_set_local()` removes the bit for the other IP version in
`ce_mask`. A missing flag inversion in the v4 part made this removal
reset all bits to 0 except the v6 one, screwing all link configuration.


thom311:
merged as cd5f6f842905b83d8c49f3eacb17273997036552

Thanks!!

PR #194, opened 2018-06-12
Add extra type tunnel types and fwmark API

TummyFish:
Added support for IPv6 GRE and IPv6 VTI tunnel types.
Added API to configure fwmark on tunnels. This required updating if_tunnel.h up to IFLA_*_FWMARK.
Fixed some bugs with using wrong define for some arrays.


lib/route/link/ip6gre.c
```diff
@@ -0,0 +1,788 @@
...
+		ip6gre->ip6gre_mask |= IP6GRE_ATTR_OKEY;
+	}
+
+	if (tb[IFLA_GRE_LOCAL]) {
+		nla_memcpy(&ip6gre->local, tb[IFLA_GRE_LOCAL], sizeof(struct in6_addr));
+		ip6gre->ip6gre_mask |= IP6GRE_ATTR_LOCAL;
+	}
+
+	if (tb[IFLA_GRE_REMOTE]) {
+		nla_memcpy(&ip6gre->remote, tb[IFLA_GRE_LOCAL], sizeof(struct in6_addr));
```
TummyFish: IFLA_GRE_LOCAL should be IFLA_GRE_REMOTE on this line

thom311: please feel invitated to fix the branch and just force-push to github. That will update the pull-request.

include/netlink/route/link/ip6gre.h
```diff
@@ -0,0 +1,64 @@
...
+	extern uint32_t rtnl_link_ip6gre_get_okey(struct rtnl_link *link);
+
+	extern int rtnl_link_ip6gre_set_local(struct rtnl_link *link, struct in6_addr *local);
+	extern int rtnl_link_ip6gre_get_local(struct rtnl_link *link, struct in6_addr *local);
+
+	extern int rtnl_link_ip6gre_set_remote(struct rtnl_link *link, struct in6_addr *remote);
+	extern int rtnl_link_ip6gre_get_remote(struct rtnl_link *link, struct in6_addr *remote);
+
+	extern int rtnl_link_ip6gre_set_ttl(struct rtnl_link *link, uint8_t ttl);
+	extern uint8_t rtnl_link_ip6gre_get_ttl(struct rtnl_link *link);
```
thom311: getter should be able to signal an error, like `-NLE_NOATTR` or `-NLE_INVAL`. For that, their signature should always return `int` type, while returning the value as pointer. Yes, that is unfortunately not done consistently, but for new API I think that is the way how it should be done. This means for example:
```C
int rtnl_link_ip6gre_get_ttl(struct rtnl_link *link, uint8_t *out_val)
{
        struct ip6gre_info *ip6gre = link->l_info;

        IS_IP6GRE_LINK_ASSERT(link);

        *out_val = ip6gre->ttl;
        return 0;
}
```

Possibly there is also:
```C
        if (!(ip6gre->ip6gre_mask & IP6GRE_ATTR_TTL))
                return -NLE_NOATTR;
```

TummyFish:
Any update on this?

TummyFish:
I re-pushed my branch a while back with the requested changes so can this be delivered?

thom311:
commit "ip6gre: Add IPv6 GRE support" got merged as 4edffbd41a42730d245823f9968b593dfcb9b6f6.
Sorry, I did not realize that it's the same patch, so the attribution is not correct.

thom311:
I rebased the remaining patches to master, and merge them. They are very good, thank you.

I am very sorry for being so non-responsive.  Thank you.

PR #196, opened 2018-07-26
Qdisc mqprio

voldymyr:
Hi, please have a look at this patch set. It adds full range of configuration options for mqprio qdisc.
For more information please have a look at: http://man7.org/linux/man-pages/man8/tc-mqprio.8.html


thom311:
thanks. Merged with a few follow-up fixes

PR #197, opened 2018-07-27
This patch allows to set chain index for tc objects

voldymyr:
In this way one can chain together a set of filters for different purposes.


thom311:
merged to master. Thanks.

PR #198, opened 2018-07-27
act-vlan

voldymyr:
This patch set adds vlan action.
For more information about vlan action please have a look at: http://man7.org/linux/man-pages/man8/tc-vlan.8.html


thom311:
Rebased to master, and applied a few fixes thereby (fix conflicts, and add `include/linux-private/linux/tc_act/tc_vlan.h` to `Makefile.am`).

Then, added two more commits on top of it, and merged to master.

Thanks!

PR #199, opened 2018-07-27
nla_ok: fix overrun in attribute iteration.

havelptess:
A detailed explanation is provided in the original Linux kernel commit that
fixes the bug: 1045b03e07d85f3545118510a587035536030c1c

Valgrind spotted the issue when the remaining was negative.
This bug was triggering application crashes.

Signed-off-by: Patrick Havelange <patrick.havelange@tessares.net>


thom311:
Merged. Thank you!

PR #200, opened 2018-07-30
cls-flower

voldymyr:
For more information about flower classifier please have a look here: 
http://man7.org/linux/man-pages/man8/tc-flower.8.html

This patch adds a subset of functions.


lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+static struct nla_policy flower_policy[TCA_FLOWER_MAX + 1] = {
+        [TCA_FLOWER_KEY_ETH_TYPE]	= { .type = NLA_U16 },
+        [TCA_FLOWER_KEY_ETH_DST]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_DST_MASK]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_SRC]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_SRC_MASK]	= { .maxlen = ETH_ALEN },
+        [TCA_FLOWER_KEY_VLAN_ID]	= { .type = NLA_U16 },
+	[TCA_FLOWER_KEY_VLAN_PRIO]	= { .type = NLA_U8 },
+        [TCA_FLOWER_KEY_IP_TOS]		= { .type = NLA_U8 },
+	[TCA_FLOWER_KEY_IP_TOS_MASK]	= { .type = NLA_U8 },
```
thom311: hi. This seems wrong. Below `flower_msg_parser` expects attributes like `TCA_FLOWER_FLAGS` (and U32 int). But here, the policy has some other fields.

voldymyr: Hi, yes it seems like I had missed to add TCA_FLOWER_FLAGS to policy, so that it can be verified. I will add it.

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+		f->cf_mask |= FLOWER_ATTR_PROTO;
+	}
+
+	if (tb[TCA_FLOWER_KEY_VLAN_ID]) {
+	        f->cf_vlan_id = nla_get_u16(tb[TCA_FLOWER_KEY_VLAN_ID]);
+		f->cf_mask |= FLOWER_ATTR_VLAN_ID;
+	}
+
+	if (tb[TCA_FLOWER_KEY_VLAN_PRIO]) {
+	        f->cf_vlan_prio = nla_get_u16(tb[TCA_FLOWER_KEY_VLAN_PRIO]);
```
thom311: the priority is u8.

include/netlink/route/cls/flower.h
```diff
@@ -0,0 +1,58 @@
...
+#endif
+
+extern int      rtnl_flower_set_proto(struct rtnl_cls *cls, uint16_t);
+extern int      rtnl_flower_get_proto(struct rtnl_cls *cls, uint16_t *);
+
+extern int	rtnl_flower_set_vlan_id(struct rtnl_cls *, uint16_t);
+extern int	rtnl_flower_get_vlan_id(struct rtnl_cls *, uint16_t *);
+
+extern int	rtnl_flower_set_vlan_prio(struct rtnl_cls *, uint16_t);
+extern int	rtnl_flower_get_vlan_prio(struct rtnl_cls *, uint16_t *);
```
thom311: seems the priority is u8. Hence, the public API should also be like that.

include/netlink-private/types.h
```diff
@@ -613,6 +613,24 @@ struct rtnl_mall
 	int              m_mask;
 };
 
+struct rtnl_flower
+{
+        uint16_t                cf_proto;
+        uint16_t                cf_vlan_id;
+        uint16_t                cf_vlan_prio;
```
thom311: prioriy seems to be uint8_t

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+	if (tb[TCA_FLOWER_POLICE]) {
+		f->cf_police = nl_data_alloc_attr(tb[TCA_FLOWER_POLICE]);
+		if (!f->cf_police)
+			return -NLE_NOMEM;
+
+		f->cf_mask |= FLOWER_ATTR_POLICE;
+	}
+
+	if (tb[TCA_FLOWER_KEY_ETH_TYPE]) {
+	        f->cf_proto = nla_get_u16(tb[TCA_FLOWER_KEY_ETH_TYPE]);
```
thom311: this and all other u16 types here are in big-endian (network) byte order. The parsed value should be corrected for that.

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+	        err = rtnl_act_fill(msg, TCA_FLOWER_ACT, f->cf_act);
+		if (err)
+		        return err;
+	}
+
+	if (f->cf_mask & FLOWER_ATTR_POLICE)
+	        NLA_PUT_DATA(msg, TCA_FLOWER_POLICE, f->cf_police);
+
+	if (f->cf_mask & FLOWER_ATTR_PROTO)
+	        NLA_PUT_U16(msg, TCA_FLOWER_KEY_ETH_TYPE, f->cf_proto);
```
thom311: this and most (all?) other U16 values need to be in big endian byte order on netlink API. As I suggested above to correct for endianness during parsing, the process must be reverted during `NLA_PUT_U16()`.

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+#define FLOWER_DSCP_MASK_MAX        0xe0
+#define FLOWER_VID_MAX              4095
+#define FLOWER_VLAN_PRIO_MAX        7
+
+static struct nla_policy flower_policy[TCA_FLOWER_MAX + 1] = {
+        [TCA_FLOWER_KEY_ETH_TYPE]	= { .type = NLA_U16 },
+        [TCA_FLOWER_KEY_ETH_DST]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_DST_MASK]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_SRC]	= { .maxlen = ETH_ALEN },
+	[TCA_FLOWER_KEY_ETH_SRC_MASK]	= { .maxlen = ETH_ALEN },
```
thom311: the MAC addresses are all `ETH_ALEN` bytes. Please also set `.minlen = ETH_ALEN`.

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+		f->cf_mask |= FLOWER_ATTR_VLAN_PRIO;
+	}
+
+	if (tb[TCA_FLOWER_KEY_VLAN_ETH_TYPE]) {
+	        f->cf_vlan_ethtype = nla_get_u16(tb[TCA_FLOWER_KEY_VLAN_ETH_TYPE]);
+		f->cf_mask |= FLOWER_ATTR_VLAN_ETH_TYPE;
+	}
+
+	if (tb[TCA_FLOWER_KEY_ETH_DST]) {
+	        f->cf_dst_mac = nl_data_alloc_attr(tb[TCA_FLOWER_KEY_ETH_DST]);
```
thom311: the MAC addresses are all fixed size of `ETH_ALEN` bytes. Maybe better to embed these fields as as arrays:
```
uint8_t cf_dst_mac[ETH_ALEN];
```
Compare to `struct rtnl_link`'s `l_addr`, which is of type `struct nl_addr`. That would be a possibility, however, here the length is always fixed, so embedding the array in the struct seems preferable to me.

Note also API like `rtnl_flower_get_dst_mac()` where the length `ETH_ALEN` is fixed in the API. If you'd ever believe the length could be any different, then this API would need to change so that the length of the MAC address can be specified/retrieved.

lib/route/cls/flower.c
```diff
@@ -0,0 +1,779 @@
...
+	        nl_dump(p, " vlan_id %u", f->cf_vlan_id);
+
+	if (f->cf_mask & FLOWER_ATTR_VLAN_PRIO)
+	        nl_dump(p, " vlan_prio %u", f->cf_vlan_prio);
+
+	if (f->cf_mask & FLOWER_ATTR_VLAN_ETH_TYPE)
+	        nl_dump(p, " vlan_ethtype %u", f->cf_vlan_ethtype);
+
+	if (f->cf_mask & FLOWER_ATTR_DST_MAC)
+	        nl_dump(p, " dst_mac %s", f->cf_dst_mac);
```
thom311: Here, these MAC addresses are `struct nl_data *` and cannot be printed this way. Above I suggested to make the fields arrays. Either way, the MAC addresses are binary and not NUL terminated. They cannot be printed this way. For example, link.c does:
```
nl_dump(p, "%s ", nl_addr2str(link->l_addr, buf, sizeof(buf)));
```
though, I don't think that the addresses should be `struct nl_addr` because their length is fixed. But something along the line.

lib/route/cls/flower.c
voldymyr: cf_proto and cf_vlan_ethtype hold values in big-eidian. Whenever user sets values, they are converted to big-endian. In this case kernel returns already a big-endian value in nla_get_u16(). The same is valid for vlan_ethtype, it's also in big-endian as kernel returns it.

lib/route/cls/flower.c
voldymyr: cf_proto and cf_vlan_ethtype are always in big-endian. No conversion is needed here.

lib/route/cls/flower.c
voldymyr: Sure, I think making mac/mask fields as arrays, as you suggested, is pretty good.

lib/route/cls/flower.c
voldymyr: Oh, I missed that. Thx.

voldymyr:
I have now updated the code according to your comments. Please have a look whenever you have time. Thanks.

thom311:
merged to master.

I squashed all the commit. Because the follow-up commits were just fixes for the first commit. It is often desirable to split a pull-request in several patches, but not if the follow-up commits only fix bugs of the earlier commits. Also, the intermediate commits did not compile.

I also dropped `TCA_FLOWER_POLICE`, otherwise it cannot build. This attribute was never merged to upstream kernel.

Also added some follow-up commits.


Thank you for the pull request!!!

PR #201, opened 2018-08-23
Fix potential memory leak in route/cls

ipronin:
Commit e5d9b828f6ec64fd77854578fbf1c33f214f3ac4 added reference grabbing for actions that are added to the filter before inserting them to the list. `rtnl_act_append()` fails when it tries to add more than `TCA_ACT_MAX_PRIO` actions to the same list. Because of that `rtnl_basic_add_action()` and `rtnl_u32_add_action()` should not increment the reference counter of the given action until it is successfully added to the filter's list.


thom311:
merged. Thank you!

PR #205, opened 2018-10-31
Add 64bit rate/ceil support for htb class

d0u9:
Htb class has already supported 64bit rate and ceil settings for times.
Now, in this patch, we grant this ability to libnl library.

### DIFF
--- a/doc/route.txt
+++ b/doc/route.txt
@@ -1843,8 +1843,8 @@ or erqual than the rate of its children.
 +
 [source,c]
 -----
-uint32_t rtnl_htb_get_rate(struct rtnl_class *class);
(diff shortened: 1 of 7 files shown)


thom311:
Hm, this changes API and ABI and we cannot do that. The only way is to add new accessors for the 64 bit values, that obsoletes the other ones.

On the other hand, `struct rtnl_ratespec` is an internal type. So its `rs_rate` field could be extended to u64...

thom311:
Merge the patch to master, with a follow-up that reverts the API change.

If you dislike any of these changes or if you see anything wrong, please comment.


Thank you for the patch!!

PR #213, opened 2019-03-27
Use nl_send_auto() instead of deprecated nl_send_auto_complete() in t…

gonapps:
Use nl_send_auto() instead of deprecated nl_send_auto_complete() in test-genl.c


thom311:
merged to master. Thank you!

PR #218, opened 2019-06-14
Add CTA_LABELS and CTA_LABELS_MASK to ctattr_type according to the ne…

xbzheng166:
Some developer might need add their own items to the end of ctattr_type, to pass their own fields to the conntrack in the kernel. If ctattr_type in libnl is not consistent with the one in linux kernel, their own added items will have different id in libnl and in kernel.


thom311:
merged to master. Thank you

PR #219, opened 2019-06-20
Add SPDX identifiers

yegorich:
Software Package Data Exchange identifiers help to detect source file
licenses and hence simplify the FOSS compliance process.

Signed-off-by: Yegor Yefremov <yegorslists@googlemail.com>

### DIFF
--- a/lib/addr.c
+++ b/lib/addr.c
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
 /*
  * lib/addr.c		Network Address
  *
--- a/lib/attr.c
+++ b/lib/attr.c
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
 /*
  * lib/attr.c		Netlink Attributes
  *
--- a/lib/cache.c
+++ b/lib/cache.c
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
 /*
  * lib/cache.c		Caching Module
  *
--- a/lib/cache_mngr.c
+++ b/lib/cache_mngr.c
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
 /*
  * lib/cache_mngr.c	Cache Manager
  *
--- a/lib/cache_mngt.c
+++ b/lib/cache_mngt.c
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
 /*
  * lib/cache_mngt.c	Cache Management
  *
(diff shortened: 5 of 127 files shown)


thom311:
This is a very valuable effort. Thank you.


Did you do this manually, or did you have a script?

What exactly was your approach when creating this patch?

thom311:
I see, the patch perfectly adds the a first line with SPDX header... so, I guess it was a script (or just as good).

That leaves the question, how did you select the files where the LGPL-2.1 license applies? Did you manually look through the files and check the license headers of all files?


I ask these question, because I'd like to understand how you did it. In particular, if it was a script, then it would be useful to verify the script (for correctness). If it was manual, then the "review" means to also check it manually -- also, the manual process is less reproducible.

yegorich:
@thom311 I've simply used `sed` to add the SPDX identifier. I've looked in some files in a directory and looked for a license definition. It could be that I've missed some.

My goal was to create a report with package licenses, copyright information etc. I use https://www.fossology.org/ to extract this information from my buildroot package archives. This was a very tedious task because fossology scans source files and tries to guess the license(s).

Now I'm trying another approach, I send patches adding SPDX identifiers. So that for the future releases I don't have to check/guess the licenses of the projects I use. This was also inspired by this [paper](https://www.linuxfoundation.org/publications/2018/03/license-scanning-compliance-programs-foss-projects/).

thom311:
Thanks.

Adding SPDX ids is certainly good, and thank you for sending a patch.

I probably will merge your patch as-is, but I would also like to get rid of the verbose license headers (the comments). But I would rather not do that manually (as it's error prone), but having a non-trivial script that recognizes a fixed set of license headers and replaces them with SPDX ids. By having it scripts, I think there is less manual labour, and also it's easier to verify then manual changes.

Anyway, thank you. Will merge ...

yegorich:
FYI: https://fsfe.org/news/2019/news-20190807-01.html

rohieb:
There a few files now which have comments identifying them as "GPL-2.0-only" in plain text, but also a "SPDX-License-Identifier: LGPL-2.1-only":

./src/idiag-socket-details.c    GPL (v2)
./src/nl-addr-add.c     GPL (v2)
./src/nl-addr-delete.c  GPL (v2)
./src/nl-addr-list.c    GPL (v2)
./src/nl-cls-add.c      GPL (v2)

I think those should have a "SPDX-License-Identifier: GPL-2.0-only" instead.

thom311:
@rohieb you are right. Fixed wrong SPDX id in 503aa5e026a60d217ce389b857ce136748bd55f1 and added missing SPDX identifiers.

These files are some command line tools. It is probably an oversight that they have the wrong license. But then it doesn't really matter that much, as they are not part of the library.

rohieb:
@thom311 Thanks! Looks good to me now.

PR #220, opened 2019-08-09
[th/add-utils-and-various-cleanup]

thom311:
I think we need some convenience macros and helper to make implementing libnl more convenient (and less prone to accidental errors). Add `__attribute__((__cleanup__(fcn)))` macros and various other helpers.

Also, don't allow VLAs and fix memory corruption in xfrm-sa setters.


bengal:
Looks good to me.

PR #221, opened 2019-08-15
Various portability fixes

michaelforney:


thom311:
A not about C features and required standard to build libnl3.

This is about compiling libnl3 itself. I am not talking about the public API (public headers) of libnl3, which of course should be very conservative w.r.t. C standards and gcc-isms. Basically, public headers should be C89.

But to compile libnl3 itself, we shouldn't target a hypothetical `--std=c99`. The goal is to compile libnl3 with real world compilers, and use features that those provide. So, if you had said that you'd like to compile libnl3 with a certain compiler (compiler version), and that failed, then the argument would be more convincing than trying to compile with `--std=c99 --pedantic` because you'd like to. I think we should not aim for that. We should see that we use features that are supported by *all* the compilers, that we want to be able to compile libnl3.

I am also in favour of enabling all kind of compiler warnings and make sure that we build without any warnings. But some warnings don't make too much sense. `--std=c99` basically turns on additional warnings, but some of them don't really make sense (for building libnl3).


I'd like to use features that are available with `--std=gnu11`, and not limit ourself to `--std=c99`. Yes, I think we should slowly updating to C11/gnu11. Note that clang is fine with most of these gcc extensions, and when they are not, then that is a good reason to not use them. But let's not pro-actively avoid features, when all compilers are actually fine with them.


The reason not to target a strict C standard is that `__attribute__(cleanup())` and `__typeof__()` are too good to miss. Also `_Generic()` is nice, which is C11 but not C99.


Of course, most changes of the branch are uncontroversial and don't take away important features. I don't mind using `__typeof__` instead of `typeof`. I am less happy about refuting `?:`, because that is a really nice extension, but well... arguably, that's not that important either -- if you have `typeof()` :) .


Thank you for the patches!!

thom311:
Hm, building with clang I get lots of
```
./include/netlink-private/netlink.h:263:20: error: unused function 'nl_write_unlock' [-Werror,-Wunused-function]
static inline void nl_write_unlock(pthread_rwlock_t *lock)
                   ^
```
warnings. This warning does seem bogus, the question is how to get rid of it without killing it entirely...

michaelforney:
Thanks for the writeup and merging the changes.

While gcc and clang are the most widely used C compilers, they aren't the only ones out there. The goal with these patches is to get closer to building with my own C compiler, [cproc](https://git.sr.ht/~mcf/cproc), which is a bit conservative with regards to GNU C extensions. The patches I sent were the uncontroversial subset of the ones I needed to build libnl-3.4.0 (the others can be found at https://github.com/michaelforney/oasis/tree/master/pkg/libnl/patch).

I understand that several of these extensions are useful, but since libnl is a dependency of wpa_supplicant, an essential part of any linux system, I think restricting their use has a lot of value. Something like `?:` may be convenient, but only slightly so, and I don't think it warrants deviating from the standard. On the other hand `__typeof__` does add some real value and is pretty widespread, so I have implemented it in cproc.

Of course, it's not up to me to decide what extensions you use in libnl, but I hope you'll keep alternative implementations in mind when making that decision.

thom311:
I agree.

In particular if you have a real compiler at hand that you'd like to use (but cannot), than that is a much stronger argument.

If that compiler is a desired target, then we should add it the the CI build on travis...

thom311:
FWIW, I think the cleanup attribute is invaluable. So much so, that if a compiler does not support that, then I'd rather say, that libnl3 doesn't support that compiler (instead of avoiding the cleanup attribute). That and `__typeof__`. Everything else, I can do without.

(after all, libnl3 is pretty much dependent on the linux kernel, as it also uses linux-kernel headers. As you are on linux already, you will have a gcc or clang around anyway, so it does not unreasonably limit portability further)

PR #222, opened 2019-08-16
Sync linux headers to 4.19.66

michaelforney:
This fixes the build with musl libc.

Additionally, several changes were made to account for changes to the
headers:

- `ip_mp_alg.h` was removed, since it was removed in linux commit torvalds/linux@e06e7c61
  (v2.6.23), and the last use of those constants was removed in libnl
  commit 535e8316.
- Uses of `TCF_META_ID_SK_ROUTE_CAPS` were updated to
  `__TCF_META_ID_SK_ROUTE_CAPS`, since it was renamed in linux commit
  torvalds/linux@e20e6940 (v3.1).
- Uses of `IF_CARRIER_DOWN` and `IF_CARRIER_UP` were replaced with their
  values, 0 and 1, since they are not in `linux/if.h` (they appear to be
  libnl-specific, added in libnl commit 3540e44b).


michaelforney:
Hmm... it looks like `uint16_t`, used for parameter `ext` in `idiagnl_send_simple` is no longer big enough to store `_INET_DIAG_ALL`. Not sure what to do about this.

thom311:
I merged the pull-request. Thank you!!

> Hmm... it looks like uint16_t, used for parameter ext in idiagnl_send_simple is no longer big enough to store _INET_DIAG_ALL. Not sure what to do about this.

Right. Internally it even only considers 8 bits, so `0xFF00` flags are ignored as well. 

How about https://github.com/thom311/libnl/commit/194069516dcd16071314c110b18c3f325eac414f ? For now I just added a comment about that. If anybody requires this to work, patch welcome.

PR #232, opened 2020-03-10
Remove VLAN from mandatory neigh comparison flags

karel-dolezal:
### Issue
The neighbor cache fails to compare objects for equality and is leaking objects stored in the neighbor hash table. The VLAN attribute is listed as mandatory to enable neighbor object comparison, but it is very often not received from linux at all.

### Testing
I'm running libnl 3.4 with this to fix the leak: [patch_for_libnl_3.4.patch.txt](https://github.com/thom311/libnl/files/4311167/patch_for_libnl_3.4.patch.txt). I've not had a chance to test with 3.5.

I've tried running kernels 4.1 and 4.19, both of them were NOT always sending the VLAN attribute, resulting in a leak, unless this patch was applied.

### Details
Remove NEIGH_ATTR_VLAN from the set of mandatory comparison flags, allowing netlink neighbor cache to release all resources when being cleaned.

Linux kernels without bugfix https://github.com/torvalds/linux/commit/f82ef3e10a870acc19fa04f80ef5877eaa26f41e do not send `NDA_VLAN` for neighbors, and thus objects of `AF_BRIDGE` family never contain the `NEIGH_ATTR_VLAN` attribute. Furthermore, the attribute is not always sent even with the above change (see https://github.com/torvalds/linux/commit/47fab41ab51a18a2e5fc4ec63f16b4c6702809b6).

The attribute is (wrongly) marked as mandatory for object comparison by `nl_object_identical()` which always returns false without it. One of the consequences is hashtable's inability to remove entries during `nl_cache_clear()`. The hashtable does not release such stored objects, which do not get freed and waste memory. Refreshing the neighbor cache amplifies this leak.

### DIFF
--- a/lib/route/neigh.c
+++ b/lib/route/neigh.c
@@ -321,10 +321,9 @@ static uint32_t neigh_id_attrs_get(struct nl_object *obj)
 	if (neigh->n_family == AF_BRIDGE) {
 		if (neigh->n_flags & NTF_SELF)
 			return (NEIGH_ATTR_LLADDR | NEIGH_ATTR_FAMILY | NEIGH_ATTR_IFINDEX |
-				       ((neigh->ce_mask & NEIGH_ATTR_DST) ? NEIGH_ATTR_DST: 0) |
-				       ((neigh->ce_mask & NEIGH_ATTR_VLAN) ? NEIGH_ATTR_VLAN : 0));
+				       ((neigh->ce_mask & NEIGH_ATTR_DST) ? NEIGH_ATTR_DST: 0);
 		else
-			return (NEIGH_ATTR_LLADDR | NEIGH_ATTR_FAMILY | NEIGH_ATTR_MASTER | NEIGH_ATTR_VLAN);
+			return (NEIGH_ATTR_LLADDR | NEIGH_ATTR_FAMILY | NEIGH_ATTR_MASTER);
 	} else
 		return neigh_obj_ops.oo_id_attrs;
 }


thom311:
>  One of the consequences is hashtable's inability to remove entries during nl_cache_clear().

That seems a separate, severe problem. I think I addressed that by ab015e109ad15df03ae66681c16eeb2af0b1cd2e.

This also should fix the general problem, where we would try to compare an object that lacks an attribute. Now, the comparison should work as one would expect: two object compare identical if (and only if) they both lack the attribute.

Does that fix the issue?

---

I think the relevant question here still is whether kernel considers `NDA_VLAN` as part of the primary key of the neighbor object. OK, kernel may not always set that property, but that alone doesn't mean much here.

For example, for routes (`RTM_NEWROUTE`) most attributes are part of the ID. Take `RTA_PREFSRC`, while it may not be present on netlink (having a default value), you can add two routes that only differ in this aspect alone.

@karel-dolezal your comment is pretty clear about asserting that `NDA_VLAN` is not at all part of the primary key of a neigh object. But most of your comment talks about that the attribute is sometimes omitted and about the leak. Can you confirm that `NDA_VLAN` has nothing to do with the ID attributes?

karel-dolezal:
Yes, from what I see I think it (https://github.com/thom311/libnl/commit/ab015e109ad15df03ae66681c16eeb2af0b1cd2e) should fix it. I agree the hash-table's inability to empty is a separate problem, but I think it was a direct consequence of the broken comparison. As long as `NEIGH_ATTR_VLAN` is no longer strictly necessary, it should be fine.

-----

Regarding `NDA_VLAN`, I have no proof but think it must be part of the primary key, because we need FDB to be able to list a separate neighbor record for all the VLANs on which a certain MAC address was seen.

So I think the `NDA_VLAN` is part of the ID, but also it is only sent when nonzero. (as seen here https://github.com/torvalds/linux/commit/47fab41ab51a18a2e5fc4ec63f16b4c6702809b6)
In other words, absence of the attribute implies no VLAN tag, which is internally stored as 0.

thom311:
Thanks for the reply.

With that explaination, I think the current code is correct, and the patch (of this merge request) is not the right solution, because `NDA_VLAN` is apparently part of the ID.

I thus close this merge-request. But please reopen or report if something is still missing.

Thank you!!

PR #233, opened 2020-03-10
xfrm: remove superfluous xfrm_usersa_id from dump request

spellingmistake:
When being used with the dump flag, the kernel dispatches the request to
xfrm_dump_sa. This function does not expect the netlink message to
contain a struct xfrm_usersa_id as data payload of the netlink message.
Instead it interprets this payload as xfrm attributes. With the current
implementation, due to the sa_id being memset to zero, it causes the
kernel to complain about '24 bytes leftover after parsing attributes
[...]'.
This patch removes payload to get rid of the kernel complaint.

### DIFF
--- a/lib/xfrm/sa.c
+++ b/lib/xfrm/sa.c
@@ -650,11 +650,7 @@ static struct nla_policy xfrm_sa_policy[XFRMA_MAX+1] = {
 
 static int xfrm_sa_request_update(struct nl_cache *c, struct nl_sock *h)
 {
-	struct xfrm_id sa_id;
-
-	memset (&sa_id, 0, sizeof (sa_id));
-	return nl_send_simple (h, XFRM_MSG_GETSA, NLM_F_DUMP,
-	                       &sa_id, sizeof (sa_id));
+	return nl_send_simple (h, XFRM_MSG_GETSA, NLM_F_DUMP, NULL, 0);
 }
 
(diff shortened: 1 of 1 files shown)


spellingmistake:
He Thomas, I just now noticed that the policy dump also adds a payload which I suspect to be superfluous (struct xfrm_userpolicy_id in that case). It does however not cause the kernel warning. After a quick look at the code, I guess it would not hurt to remove that payload, too.

thom311:
Thanks Thomas, I merged your patch.

> I just now noticed that the policy dump also adds a payload which I suspect to be superfluous (struct xfrm_userpolicy_id in that case). It does however not cause the kernel warning. After a quick look at the code, I guess it would not hurt to remove that payload, too.

Hm. I think I see what you mean. But I have no experience with xfrm, and I would rely on you knowing what you do :) . Would you send a follow up patch? Thank you!

spellingmistake:
>  I would rely on you knowing what you do :)
I do, but when people rely on me, I start having doubts. Anyhow, the previous patch got rid of the warning and the functionality is still intact, I checked that. I also checked the kernel sources and figured, polices don't require that parameter either, but they don't cause a warning.
I'll put that modification on my bucket list and when I have time to test it, I'll get back at you.

thom311:
@spellingmistake thank you!!

PR #234, opened 2020-03-23
Add RTM_MDB operations

rubensfig:
The Linux kernel has a notification system via Netlink that reports the
changes in the multicast database over the RTNLGRP_MDB multicast socket.
As such notifications such as RTM_NEWMDB, RTM_GETMDB can be processed in
userspace.
libnl must support the capability of processing these messages, e.g. RTM_GETMDB.
This commit adds a new rtnl_mdb object as well as a route/mdb cache,
in order to expose the contents of the multicast database.

### DIFF
--- a/lib/route/mdb.c
+++ b/lib/route/mdb.c
@@ -1,0 +1,452 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/*
+ * lib/route/mdb.c		Multicast Database
+ */
+
+#include <netlink-private/netlink.h>
+#include <netlink/netlink.h>
+#include <netlink/route/mdb.h>
+#include <netlink/utils.h>
+#include <linux/if_bridge.h>
+
+/** @cond SKIP */
+#define MDB_ATTR_IFINDEX         0x000001
+#define MDB_ATTR_PORT            0x000002
+#define MDB_ATTR_ADDRESS         0x000004
+#define MDB_ATTR_VID             0x000008
+
+static struct nl_cache_ops rtnl_mdb_ops;
+static struct nl_object_ops mdb_obj_ops;
+/** @endcond */
+
+static void mdb_constructor(struct nl_object *obj)
+{
+	struct rtnl_mdb *_mdb = (struct rtnl_mdb *) obj;
+
+	nl_init_list_head(&_mdb->mdb_entry_list);
+}
+
+static void mdb_entry_free_data(struct rtnl_mdb_entry *entry)
+{
+	nl_addr_put(entry->addr);
+	free(entry);
+}
+
+static void mdb_free_data(struct nl_object *obj)
+{
+	struct rtnl_mdb *mdb = (struct rtnl_mdb *) obj;
+	struct rtnl_mdb_entry *mdb_entry;
+
+	nl_list_for_each_entry(mdb_entry, &mdb->mdb_entry_list, mdb_list)
+	    mdb_entry_free_data(mdb_entry);
+}
+
+static uint64_t mdb_entry_compare(struct nl_object *_a, struct nl_object *_b,
+				  uint64_t attrs, int flags)
+{
+	uint64_t diff = 0;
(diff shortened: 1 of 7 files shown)


libnl-route-3.sym
```diff
@@ -1149,4 +1149,15 @@ global:
...
+	rtnl_mdb_alloc_cache_flags;
+	rtnl_mdb_get_ifindex;
+	rtnl_mdb_get_family;
+	rtnl_mdb_add_entry;
+	rtnl_mdb_foreach_entry;
+  rtnl_mdb_entry_get_ifindex;
+  rtnl_mdb_entry_get_vid;
+  rtnl_mdb_entry_get_state;
+  rtnl_mdb_entry_get_addr;
+  rtnl_mdb_entry_get_proto;
```
thom311, line 1162: please indent the entires following the current style. Also, sort the lines asciibetically.

More important, a linker symbol version must not be modified after there is a stable release. That means, don't add entries to the existing `libnl_3_5` section, instead create a new `libnl_3_6` section.

lib/route/mdb.c
```diff
@@ -0,0 +1,289 @@
...
+	.co_msgtypes		= {
+					{ RTM_NEWMDB, NL_ACT_NEW, "new"},
+					{ RTM_DELMDB, NL_ACT_DEL, "del"},
+					{ RTM_GETMDB, NL_ACT_GET, "get"},
+					END_OF_MSGTYPES_LIST,
+				  },
+	.co_protocol		= NETLINK_ROUTE,
+  .co_groups		= mdb_groups,
+  .co_request_update	= mdb_request_update,
+  .co_msg_parser = mdb_msg_parser,
```
thom311, line 327: indentation.

In the whole patch, please be mindful that. No trailing whitespace, and indent so that it looks good with 8 space tabs.

lib/route/mdb.c
```diff
@@ -0,0 +1,289 @@
+/*
+ * lib/route/mdb.c		Multicast Database
```
thom311, line 2: for new files, please add a `/* SPDX-License-Identifier: LGPL-2.1-only */` line as the first line, and don't add the redundant license text comments.

The copyright comments only apply if you copied and adjusted the code. I guess, you did actually take the existing code and adjusted it. In that case, it's correct to preserve the copyright comments.

lib/route/mdb.c
```diff
@@ -0,0 +1,289 @@
...
+      entry = nla_data(entry_attr[MDBA_MDB_ENTRY_INFO]);
+
+      _nentry->ifindex = entry->ifindex;
+      _nentry->vid = entry->vid;
+      _nentry->state = entry->state;
+
+      _nentry->proto = ntohs(entry->addr.proto);
+
+      if (_nentry->proto == ETH_P_IP) {
+          char _mcast_address[INET_ADDRSTRLEN]= "";
```
thom311, line 140: Let's not use leading `_` for local variables. That is not following the current style (and does not seem a common practice to me).

lib/route/mdb.c
```diff
@@ -0,0 +1,289 @@
...
+
+          inet_ntop(AF_INET, &entry->addr.u.ip4, _mcast_address, INET_ADDRSTRLEN);
+          nl_addr_parse(_mcast_address, AF_INET, &_nentry->addr);
```
thom311, line 143: yoiu already have the address in binary, no need to convert it to string, and then use `nl_addr_parse()`. Use `nl_addr_set_binary_addr()` and `nl_addr_set_family()`.

rubensfig:
@thom311 Could you please take a look at this again?

rubensfig:
Since the previous review, I had to add:
> Support for oo_free_data, to correctly cleanup the objects
> General rework of how stuff gets added to the caches
> Improved parsing of the netlink message
> General memory hardening of the code

I think this patch is currently ready to be re

thom311:
sorry for the extremely long delay for the response.

I merged your patch, with a few follow-up patches that I thought would be good. If you find any problem with them, we can fix that.

Thank you for the patch and work!!

PR #237, opened 2020-04-14
route/link: Move LINK_ATTR_IFNAME assignment to a proper location

voldymyr:
It should be set right where link name is set.

Signed-off-by: Greger Wrang <greger.wrang@westermo.se>


thom311:
merged to master. Thank you!!

PR #238, opened 2020-04-14
route/link: Check for null pointer in macvlan

voldymyr:
In cases where link->l_info is not set, a null-ptr-exception
will be invoked.


thom311:
merged to master. Thanks!

PR #240, opened 2020-04-14
route/addr: add support for setting/getting address alias

voldymyr:


include/linux-private/linux/if_addr.h
```diff
@@ -34,6 +34,7 @@ enum {
 	IFA_MULTICAST,
 	IFA_FLAGS,
 	IFA_RT_PRIORITY,  /* u32, priority/metric for prefix route */
+	IFA_ALIAS,
```
thom311, line 37: these files in `include/linux-private` shall only be imported as a whole from recent kernels. They must not deviate from upstream kernel.

Also, where is `IFA_ALIAS` from? I don't find it in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/

lib/route/addr.c
```diff
@@ -840,6 +851,25 @@ int rtnl_addr_delete(struct nl_sock *sk, struct rtnl_addr *addr, int flags)
  * @{
  */
 
+int rtnl_addr_set_alias(struct rtnl_addr *addr, const char *alias)
+{
+	if (strlen(alias) > sizeof(addr->a_alias) - 1)
+		return -NLE_RANGE;
+
+	strcpy(addr->a_alias, alias);
```
thom311, line 859: maybe use `_nl_strncpy_assert()`, because that takes care to zero out the trailing bytes.

lib/route/addr.c
```diff
@@ -840,6 +851,25 @@ int rtnl_addr_delete(struct nl_sock *sk, struct rtnl_addr *addr, int flags)
  * @{
  */
 
+int rtnl_addr_set_alias(struct rtnl_addr *addr, const char *alias)
+{
+	if (strlen(alias) > sizeof(addr->a_alias) - 1)
```
thom311, line 856: since the getter can return `%NULL`, possibly the setter also should accept %NULL, to make the value unset. Yes, that is not done for other attributes, which I find unfortunate :(

libnl-route-3.sym
```diff
@@ -1063,6 +1063,8 @@ global:
 
 libnl_3_5 {
 global:
+	rtnl_addr_get_alias;
+	rtnl_addr_set_alias;
```
thom311, line 1067: needs to move to linker section `libnl_3_6`.

include/linux-private/linux/if_addr.h
voldymyr: _**IFA_ALIAS**_ flag does not exist in upstream kernel yet. I'm wondering now what is the policy around this kind of matters? Should we first send _IFA_ALIAS_ patch to linux kernel, and when it gets accepted, make a new pull request to libnl? If that is the case then I will close this merge-request and open it later at an appropriate time.

thom311: Yes, first the feature needs to be added to kernel.

in particular, the kernel headers (that defines magic numbers like `IFA_ALIAS`) are the stable API of the kernel. We must not invent a different API, that is not yet accepted by kernel.

Point in case, this adds `IFA_ALIAS` with numeric value 10. In the meantime, upstream kernel added `IFA_TARGET_NETNSID` with that number.

Closing this pull request. We can only support features of the upstream, officially released kernel.

thom311:
closing, as `IFA_ALIAS` is not (yet?) an upstream kernel feature. Please reopen, if that changes.

troglobit:
Hi @thom311 (and @voldymyr :), we got `IFA_PROTO` upstream instead, same idea to support labeling the originator of an IP address (static/link-local/dhcp).  It's an integer instead of a string and similar to the proto for routes.

I will submit a patch for this later this week (going through our backlog at https://github.com/westermo/libnl/commits/wmo/v3.5.0).

PR #242, opened 2020-04-15
route/route: support FIB lookups using rtnl

voldymyr:
Using the flnl_* family of functions to perform FIB lookups is rather
limited. In particular, there seems to be no way of resolving the
nexthop. By hooking into RTM_GETROUTE, a regular rtnl route object is
returned instead.

### DIFF
--- a/lib/route/route.c
+++ b/lib/route/route.c
@@ -131,6 +131,33 @@ int rtnl_route_build_add_request(struct rtnl_route *tmpl, int flags,
 			       result);
 }
 
+int rtnl_route_lookup(struct nl_sock *sk, struct nl_addr *dst,
+		      struct rtnl_route **result)
+{
+	struct nl_msg *msg;
+	struct rtnl_route *tmpl;
+	struct nl_object *obj;
+	int err;
+
+	tmpl = rtnl_route_alloc();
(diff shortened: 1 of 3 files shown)


lib/route/route.c
```diff
@@ -131,6 +131,33 @@ int rtnl_route_build_add_request(struct rtnl_route *tmpl, int flags,
...
+	rtnl_route_put(tmpl);
+	if (err < 0)
+		return err;
+
+	err = nl_send_auto(sk, msg);
+	nlmsg_free(msg);
+	if (err < 0)
+		return err;
+
+	if ((err = nl_pickup(sk, route_msg_parser, &obj)) < 0)
```
thom311, line 154: if you send a `RTM_GETROUTE` request, then you only should honor the responses with the matching `nlmsg_seq` number. The socket could have other messages pending, for example events. Isn't that a problem?

Well, maybe it actually is not a problem, because I don't think you can meaningfully use such an API together with getting unexpected events on the socket. If events (other messages) are received, then the user needs to read each message themselves and associate them with the right request.

OK.

thom311:
merged to master.

Thank you!!

PR #243, opened 2020-04-15
route/cls: allow fetching of classifiers from cache

voldymyr:
API:
	rtnl_cls_get()
	rtnl_cls_get_by_prio()

These functions search for classifiers in classifier cache.

### DIFF
--- a/lib/route/cls.c
+++ b/lib/route/cls.c
@@ -389,8 +389,9 @@ struct rtnl_cls *rtnl_cls_get(struct nl_cache *cache, int ifindex, uint32_t pare
 		return NULL;
 
 	nl_list_for_each_entry(cls, &cache->c_items, ce_list) {
-		if ((cls->c_parent == parent) && (cls->c_ifindex == ifindex)
-		    && (cls->c_handle == handle)) {
+		if ((cls->c_parent == parent) &&
+		    (cls->c_ifindex == ifindex)&&
+		    (cls->c_handle == handle)) {
 			nl_object_get((struct nl_object *) cls);
 			return cls;
 		}
@@ -424,8 +425,9 @@ struct rtnl_cls *rtnl_cls_get_by_prio(struct nl_cache *cache, int ifindex,
 		return NULL;
 
 	nl_list_for_each_entry(cls, &cache->c_items, ce_list) {
-		if ((cls->c_parent == parent) && (cls->c_ifindex == ifindex)
-		    && (cls->c_prio == prio)) {
+		if ((cls->c_parent == parent) &&
+		    (cls->c_ifindex == ifindex) &&
+		    (cls->c_prio == prio)) {
 			nl_object_get((struct nl_object *) cls);
 			return cls;
 		}
--- a/lib/route/cls.c
+++ b/lib/route/cls.c
@@ -364,6 +364,76 @@ void rtnl_cls_cache_set_tc_params(struct nl_cache *cache,
 	cache->c_iarg2 = parent;
 }
 
+/**
+ * Search classifier by interface index, parent and handle
+ * @arg cache           Classifier cache
+ * @arg ifindex         Interface index
+ * @arg parent          Parent
+ * @arg handle          Handle
+ *
+ * Searches a classifier cache previously allocated with rtnl_cls_alloc_cache()
+ * and searches for a classifier matching the interface index, parent
+ * and handle.
+ *
+ * The reference counter is incremented before returning the classifier,
+ * therefore the reference must be given back with rtnl_cls_put() after usage.
+ *
+ * @return Classifier or NULL if no match was found.
+ */
+struct rtnl_cls *rtnl_cls_get(struct nl_cache *cache, int ifindex, uint32_t parent,
+			      uint32_t handle)
+{
+	struct rtnl_cls *cls;
+
+	if (cache->c_ops != &rtnl_cls_ops)
+		return NULL;
+
+	nl_list_for_each_entry(cls, &cache->c_items, ce_list) {
+		if ((cls->c_parent == parent) && (cls->c_ifindex == ifindex)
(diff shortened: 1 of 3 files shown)


libnl-route-3.sym
```diff
@@ -1065,6 +1065,8 @@ libnl_3_5 {
 global:
 	rtnl_class_get_by_parent;
 	rtnl_cls_cache_set_tc_params;
+	rtnl_cls_get;
+	rtnl_cls_get_by_prio;
```
thom311, line 1069: needs new linker section `libnl_3_6`.

lib/route/cls.c
```diff
@@ -364,6 +364,76 @@ void rtnl_cls_cache_set_tc_params(struct nl_cache *cache,
...
+			      uint32_t handle)
+{
+	struct rtnl_cls *cls;
+
+	if (cache->c_ops != &rtnl_cls_ops)
+		return NULL;
+
+	nl_list_for_each_entry(cls, &cache->c_items, ce_list) {
+		if ((cls->c_parent == parent) && (cls->c_ifindex == ifindex)
+		    && (cls->c_handle == handle)) {
```
thom311, line 393: could we have one condition per line? This is hard to read for me

lib/route/cls.c
```diff
@@ -364,6 +364,76 @@ void rtnl_cls_cache_set_tc_params(struct nl_cache *cache,
...
+{
+	struct rtnl_cls *cls;
+
+	if (cache->c_ops != &rtnl_cls_ops)
+		return NULL;
+
+	nl_list_for_each_entry(cls, &cache->c_items, ce_list) {
+		if ((cls->c_parent == parent) && (cls->c_ifindex == ifindex)
+		    && (cls->c_prio == prio)) {
+			nl_object_get((struct nl_object *) cls);
```
thom311, line 429: what do you think about having the new API not increase the ref-count?

lib/route/cls.c
```diff
@@ -364,6 +364,76 @@ void rtnl_cls_cache_set_tc_params(struct nl_cache *cache,
...
+ * and searches for a classifier matching the interface index, parent
+ * and prio.
+ *
+ * The reference counter is incremented before returning the classifier,
+ * therefore the reference must be given back with rtnl_cls_put() after usage.
+ *
+ * @return Classifier or NULL if no match was found.
+ */
+struct rtnl_cls *rtnl_cls_get_by_prio(struct nl_cache *cache, int ifindex,
+				      uint32_t parent, uint16_t prio)
```
thom311, line 419: this patch adds two new API for things that are already possible (by the user implementing the functions themself).

Can you give some arguments why this is necessary and a good idea? Also, now we add two new functions, will this suffice? Is by-prio somehow specially interesting? Also, for a cache function called "get", I would expect that it can only find one possible value. Is the (ifindex,parent,prio) tuple in the cache unique? If not, shouldn't it be called "find_first_by_prio()"?

Also, the code comment should indicate that this that a search over all elements and is `O(n)`.

lib/route/cls.c
voldymyr: Good question. I've been browsing through the code and it seems like a common practice to increase the reference counter. However, I do not see any major hazards to removing _nl_obj_get()_ from the new API, since it would make things somewhat simpler: the user wouldn't have to bother calling respective _*_put()_ function for his object.
**The common scenario is this I believe:**
1. cache_alloc()
2. fetch object from cache
3. use object
4. cache_free()
Destroying the cache would invoke _nl_obj_put() -> nl_obj_free()_ chain for each object.

**There's one corner case popping out as the result of such change though:**
_1. cache_alloc()
2. fetch object from cache
3. cache_free()
4.let's use our object -> ooops, it is nullified._

Having the refcount incremented every time we fetch object from cache, would never leave us with nullified object.

Which way to go is a matter of preference. What do you think?

lib/route/cls.c
voldymyr: It is indeed possible by the user to implement those functions, however, the beauty of any library is in the richness of its API functionality.

These two functions make things simpler for users, since:
- no need to implement them, they are there already
- use them in multiple applications, without having to duplicate code (this is primary reason for why I'm adding this patch)
- it seems to me that they are sort of _'natural extension'_ to what already is implemented

As for whether these two functions will suffice - I believe yes. Some classifiers allow to set unique handle when adding them (u32), others not (these end up having the same handle '0x1'). _Priority_ or _preference_ is possible to set for all classifiers: it must be unique, and is a good way to fetch wanted classifier from cache.

I agree with you that the naming of functions could be improved. How about:

- _rtnl_cls_find_by_handle()_: find cls in cache by unique handle (if this cls has a unique one)
- _rtnl_cls_find_by_prio()_: find cls in cache by priority (all classifiers have unique priority)

lib/route/cls.c
voldymyr: I agree.

libnl-route-3.sym
voldymyr: moving this to new section.

lib/route/cls.c
thom311: > Which way to go is a matter of preference.

Yes, it's a matter of preference. I think it is wrong to increment the reference counter in this case, because the cache is not thread-safe. So that means, if we would return the object without a new reference, then nobody (no other thread) could remove the object while we have it. That means, it's not necessary, and make handling only more cumbersome, because you basically always need to return the reference.

However, much of existing code does return a reference. Consistency is also important.

Both approaches thus have up and downsides, I think I am fine with both (and with the latest version of the patch).

lib/route/cls.c
thom311: makes sense to add the API.

I rename the function from "get" to "find".

thom311:
merged to master.

Thank you!!

PR #244, opened 2020-04-15
route/qdisc: allow fetching qdiscs by their kind

voldymyr:
API:
	rtnl_qdisc_get_by_kind()

This function allows getting qdisc based on
its kind, i.e. tbf, htb, cbq, etc.

Signed-off-by: Volodymyr Bendiuga <volodymyr.bendiuga@westermo.se>

### DIFF
--- a/lib/route/qdisc.c
+++ b/lib/route/qdisc.c
@@ -404,6 +404,38 @@ struct rtnl_qdisc *rtnl_qdisc_get_by_parent(struct nl_cache *cache,
 }
 
 /**
+ * Search qdisc by kind
+ * @arg cache		Qdisc cache
+ * @arg ifindex		Interface index
+ * @arg kind		Qdisc kind (tbf, htb, cbq, etc)
+ *
+ * Searches a qdisc cache previously allocated with rtnl_qdisc_alloc_cache()
+ * and searches for a qdisc matching the interface index and kind.
+ *
+ * The reference counter is incremented before returning the qdisc, therefore
+ * the reference must be given back with rtnl_qdisc_put() after usage.
+ *
+ * @return pointer to qdisc inside the cache or NULL if no match was found.
+ */
+struct rtnl_qdisc *rtnl_qdisc_get_by_kind(struct nl_cache *cache,
+					    int ifindex, char *kind)
+{
+	struct rtnl_qdisc *q;
+
+	if (cache->c_ops != &rtnl_qdisc_ops)
+		return NULL;
+
+	nl_list_for_each_entry(q, &cache->c_items, ce_list) {
+		if ((q->q_ifindex == ifindex) && (!strcmp(q->q_kind, kind))) {
+			nl_object_get((struct nl_object *) q);
+			return q;
+		}
+	}
+
+	return NULL;
+}
+
+/**
  * Search qdisc by interface index and handle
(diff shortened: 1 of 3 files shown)


libnl-route-3.sym
```diff
@@ -1111,6 +1111,7 @@ global:
 	rtnl_neigh_get_master;
 	rtnl_neigh_set_master;
 	rtnl_netem_set_delay_distribution_data;
+	rtnl_qdisc_get_by_kind;
```
thom311, line 1114: we must never modify existing section in the `.sym` file, after they are released. That would completely mess up symbol versioning.

This needs to go to a new section `libnl_3_6`.

lib/route/qdisc.c
```diff
@@ -403,6 +403,38 @@ struct rtnl_qdisc *rtnl_qdisc_get_by_parent(struct nl_cache *cache,
...
+struct rtnl_qdisc *rtnl_qdisc_get_by_kind(struct nl_cache *cache,
+					    int ifindex, char *kind)
+{
+	struct rtnl_qdisc *q;
+
+	if (cache->c_ops != &rtnl_qdisc_ops)
+		return NULL;
+
+	nl_list_for_each_entry(q, &cache->c_items, ce_list) {
+		if (!strncmp(q->q_kind, kind, TCKINDSIZ) && q->q_ifindex == ifindex) {
```
thom311, line 429: various other places already require that `q->q_kind` is a NUL terminated string. Also, it seems resonable to require the caller to pass `kind` as a NUL terminated string. Just use `strcmp()`.

thom311: can you do the cheaper check first (for the ifindex)?

lib/route/qdisc.c
```diff
@@ -403,6 +403,38 @@ struct rtnl_qdisc *rtnl_qdisc_get_by_parent(struct nl_cache *cache,
...
+ * @arg ifindex		Interface index
+ * @arg kind		Qdisc kind (tbf, htb, cbq, etc)
+ *
+ * Searches a qdisc cache previously allocated with rtnl_qdisc_alloc_cache()
+ * and searches for a qdisc matching the interface index and kind.
+ *
+ * The reference counter is incremented before returning the qdisc, therefore
+ * the reference must be given back with rtnl_qdisc_put() after usage.
+ *
+ * @return pointer to qdisc inside the cache or NULL if no match was found.
```
thom311, line 418: I am not very happy about simple lookup functions increasing the reference count. On the other hand, that is consistent with `rtnl_qdisc_get_by_parent()`, so I guess it makes sense.

libnl-route-3.sym
voldymyr: Moved to new _libnl_3_6_ section.

lib/route/qdisc.c
voldymyr: I've reworked this function according to your comments.

lib/route/qdisc.c
voldymyr: I understand your reasoning, and leaving this line as it is.

voldymyr:
I've updated the pull-request branch according to your comments. Could you please have a look at it?

thom311:
merged to master.

thank you for the patch.

I apologize, that it took me so long to get this done.

PR #245, opened 2020-04-15
route/cls: U32 classifier bugfixes

voldymyr:
Spit the function into 2 parts:
 - where first being responsible for memory allocation/cloning
 - and second for populating the destination object

Make proper cloning of 'action' attribute. For now, only the
first action in chain is included in the clone.

Signed-off-by: Magnus Öberg <magnus.oberg@westermo.se>
Signed-off-by: Volodymyr Bendiuga <volodymyr.bendiuga@westermo.se>

### DIFF
--- a/lib/route/cls/u32.c
+++ b/lib/route/cls/u32.c
@@ -28,6 +28,8 @@
 #include <netlink/route/cls/u32.h>
 #include <netlink/route/action.h>
 
+#include "netlink-private/utils.h"
+
 /** @cond SKIP */
 #define U32_ATTR_DIVISOR      0x001
 #define U32_ATTR_HASH         0x002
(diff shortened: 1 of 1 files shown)


lib/route/cls/u32.c
```diff
@@ -553,6 +571,7 @@ struct rtnl_act* rtnl_u32_get_action(struct rtnl_cls *cls)
     if (!(u->cu_mask & U32_ATTR_ACTION))
         return NULL;
 
+    rtnl_act_get(u->cu_act);
```
thom311, line 574: this part should be a separate patch.

Also, it is quite a change in behavior, and will result in leaks for previously correct programs. This cannot be done. Instead, I think you can only document this behavior (of not increasing the ref count).

also, I don't understand the comment "Fix rtnl_u32_get_action() so it no longer does a raw memcpy of the action object.". `rtnl_u32_get_action()` returns the pointer value, why do you talk about raw memcpy here?

lib/route/cls/u32.c
```diff
@@ -197,6 +197,24 @@ static int u32_clone(void *_dst, void *_src)
...
+		nl_init_list_head(&dst->cu_act->ce_list);
+
+		if (src->cu_act->c_opts && !(dst->cu_act->c_opts = nl_data_clone(src->cu_act->c_opts)))
+			return -NLE_NOMEM;
+
+		if (src->cu_act->c_xstats && !(dst->cu_act->c_xstats = nl_data_clone(src->cu_act->c_xstats)))
+			return -NLE_NOMEM;
+
+		if (src->cu_act->c_subdata && !(dst->cu_act->c_subdata = nl_data_clone(src->cu_act->c_subdata)))
+			return -NLE_NOMEM;
```
thom311, line 211: all these return statement still leave the instance in an invalid state. That must be properly handled.

I would suggest to use the new `_nl_auto` macros and first clone the fields all (assigning them to `__attribute__((cleanup))` variables  on the stack), before transferring ownership at once (where we know it's no longer going to fail).

lib/route/cls/u32.c
voldymyr: Alright. I'll remove this line from current patch, as it will definitely result in leaks for existing programs.

lib/route/cls/u32.c
voldymyr: I see your point. I've reworked this function now. Please have a look when you've got time.

thom311:
merged.

Sorry for the extremely long delay.


There was a problem with the patch, that `_nl_auto_free` is not the suitable cleanup macro. It needs `_nl_auto_nl_data`.

Also, the patch is only correct after doing c027e54ffed247358b152f5330b5dc62c821024f. But that mistake was there before too.

Thanks!!

PR #247, opened 2020-04-17
Update handlers.h

vinitamaloo:


thom311:
merged to master, thanks for sending a patch!

I however reworded the commit message, as indicated in the merged patch.

PR #249, opened 2020-04-20
xfrm: remove superfluous xfrm_userpolicy_id from dump request

spellingmistake:
Analogous to the dump request for states this data structure is
unnecessary for policy dumps, too. Unlike with states it does however
not create an error message.


spellingmistake:
Hi Thomas,
my tests showed no missing policies when built with the patch.

thom311:
Note: this is a follow up to https://github.com/thom311/libnl/pull/233

Patch merged to master. Thanks Thomas!!

PR #255, opened 2020-07-17
route/link: Add IPv6 GRE support

sparlane:
This patch adds support for IPv6 GRE tunneling
that uses the ip6_gre kernel module.


thom311:
merged.

Thank you for the patch!!

thom311:
this patch is actually the same as commit "ip6gre: Add IPv6 GRE support " on #194.

PR #259, opened 2020-08-05
route/netem: fix compilation with uclibc-ng & use PATH_MAX

commodo:
On uclibc-ng, the NAME_MAX & PATH_MAX are declared in dirent.h.

NAME_MAX is the limit of a file. PATH_MAX is the limit of an absolute path,
which includes NAME_MAX.
It's a good idea to use PATH_MAX.

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@analog.com>


commodo:
Following comment https://github.com/strace/strace/pull/145#issuecomment-670019931   from here, I will close this and fix it in uclibc-ng

Thanks and apologies for the noise
Alex

PR #274, opened 2021-03-25
Bugfixed CAN support. Added CAN FD support.

pugo:
- Don't use internal bit mask constants in NLA_PUT:s. That caused most set-functions to fail.
 - Added CAN FD support analog to the normal CAN support.

### DIFF
--- a/include/netlink/route/link/can.h
+++ b/include/netlink/route/link/can.h
@@ -51,6 +51,19 @@ extern int rtnl_link_can_get_ctrlmode(struct rtnl_link *, uint32_t *);
 extern int rtnl_link_can_set_ctrlmode(struct rtnl_link *, uint32_t);
 extern int rtnl_link_can_unset_ctrlmode(struct rtnl_link *, uint32_t);
 
+extern int rtnl_link_can_get_data_bt_const(struct rtnl_link *,
+														 struct can_bittiming_const *);
+extern int rtnl_link_can_get_data_bittiming(struct rtnl_link *,
+														  struct can_bittiming *);
+extern int rtnl_link_can_set_data_bittiming(struct rtnl_link *,
+														  struct can_bittiming *);
+
+extern int rtnl_link_can_get_data_bitrate(struct rtnl_link *, uint32_t *);
+extern int rtnl_link_can_set_data_bitrate(struct rtnl_link *, uint32_t);
+
+extern int rtnl_link_can_get_data_sample_point(struct rtnl_link *, uint32_t *);
+extern int rtnl_link_can_set_data_sample_point(struct rtnl_link *, uint32_t);
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/route/link/can.c
+++ b/lib/route/link/can.c
@@ -28,25 +28,29 @@
 #include <linux/can/netlink.h>
 
 /** @cond SKIP */
-#define CAN_HAS_BITTIMING		(1<<0)
-#define CAN_HAS_BITTIMING_CONST		(1<<1)
-#define CAN_HAS_CLOCK			(1<<2)
-#define CAN_HAS_STATE			(1<<3)
-#define CAN_HAS_CTRLMODE		(1<<4)
-#define CAN_HAS_RESTART_MS		(1<<5)
-#define CAN_HAS_RESTART			(1<<6)
-#define CAN_HAS_BERR_COUNTER		(1<<7)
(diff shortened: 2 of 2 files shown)


pugo:
The first part of this PR solves a bug in `can_put_attrs()` where the internal bit mask values `CAN_HAS_*` from `lib/route/link/can.c` were used as arguments in the Netlink calls. The expected arguments are the `IFLA_CAN_*` enum values form `include/linux-private/linux/can/netlink.h`. This bug means that only some attributes were possible to set, in those cases where the bitmask values (`CAN_HAS_*`) happened to be equal to the enum values (`IFLA_CAN_*`). This was true for just a few.

The fix to above was to change `can_put_attrs()` to use the correct argument values (`IFLA_CAN_*`).

The second part of this PR is to add support for CAN FD. This is done by adding functions analog to the CAN (non FD) counterparts. The new functions have similar names to the non FD counterparts, but with the word data added. The new functions allows configuration of data bitrate, bittiming and sample point for CAN FD.

pugo:
I still suggest that this should be merged, as it solves a CAN related bug and enabled support for CAN FD.

If there is anything I can do to help and to speed up the process, please don't hesitate to tell me!

thom311:
Hi. Thanks for the patch!!

I split the patch, and merged the first part to `master`.

About the remaining part, how about https://gist.github.com/thom311/62ee22490e280abe6110edcbb89c4d75 ?

Also, I'll comment on the patch...

include/netlink/route/link/can.h
```diff
@@ -51,6 +51,19 @@ extern int rtnl_link_can_get_ctrlmode(struct rtnl_link *, uint32_t *);
 extern int rtnl_link_can_set_ctrlmode(struct rtnl_link *, uint32_t);
 extern int rtnl_link_can_unset_ctrlmode(struct rtnl_link *, uint32_t);
 
+extern int rtnl_link_can_get_data_bt_const(struct rtnl_link *,
```
thom311, line 54: there is a getter. Why is there not setter?

Below we see
```
+                 NLA_PUT(msg, IFLA_CAN_DATA_BITTIMING_CONST, sizeof(ci->ci_data_bittiming_const),
+                         &ci->ci_data_bittiming_const);
```
so it seems relevant to also set those parts. Or how is this supposed to work?

thom311: the name `rtnl_link_can_get_data_bt_const()` seems inconsistent. I think it should be `rtnl_link_can_get_data_bittiming_const()`?

include/netlink/route/link/can.h
```diff
@@ -51,6 +51,19 @@ extern int rtnl_link_can_get_ctrlmode(struct rtnl_link *, uint32_t *);
...
+extern int rtnl_link_can_get_data_bittiming(struct rtnl_link *,
+														  struct can_bittiming *);
+extern int rtnl_link_can_set_data_bittiming(struct rtnl_link *,
+														  struct can_bittiming *);
+
+extern int rtnl_link_can_get_data_bitrate(struct rtnl_link *, uint32_t *);
+extern int rtnl_link_can_set_data_bitrate(struct rtnl_link *, uint32_t);
+
+extern int rtnl_link_can_get_data_sample_point(struct rtnl_link *, uint32_t *);
+extern int rtnl_link_can_set_data_sample_point(struct rtnl_link *, uint32_t);
```
thom311, line 65: why are there here a few getters/setters for individual fields, when above we have the "get-data/set-data" to set them all? Aren't they redundant?

And about having "get-data/set-data", it requires that `struct can_bittiming(_const)` doesn't ever change the struct layout and size. I guess, that is guaranteed by kernel user space API?

include/netlink/route/link/can.h
thom311: I'll rename `rtnl_link_can_get_data_bt_const()` to `rtnl_link_can_get_data_bittiming_const()`.

I'll also add `rtnl_link_can_set_data_bittiming_const()`.

include/netlink/route/link/can.h
thom311: I'll drop those accessors (at least for now).

thom311:
I merged also the second part of the commit. I followed it up with additional commits. Hope that is ok.

Sorry for the long delay to process this. I think it's fine now.

Merged. Closing.

thom311:
thank you for the patch!

Issue #271, opened 2021-02-22
nl_cache_refill() cause removing IPv6 ECMP (multipath) routes


t0mmmy90:
Hi,

I'm using libnl 3.5.0 and I've noticed that after refilling the cache by calling nl_cache_refill() my IPv6 ECMP routes become normal ones (with only 1 next hop).

In my code I'm checking routes updates using nl_cache_mngr_poll() but after a long time without any update I'm doing the cache refill using nl_cache_refill() to be sure I'm not missing any route.
Recently I've noticed that after refill my default IPv6 route which had 2 next hops now has only 1 but in kernel it still has 2 (I'm checking with ip r s). I've checked it with the nl-route-list and it also shows 1 next hop.

It seems that this is caused by nl_cache_refill() calling internally nl_cache_pickup() instead of nl_cache_pickup_checkdup() which performs route_update added here https://lists.infradead.org/pipermail/libnl/2012-December/000836.html.

After locally modifying netlink_cache_refill() the issue is fixed (also nl-route-list shows correct number of next hops) but I'm sure there must be a reason why nl_cache_pickup() was used.

If there is any solution/workaround for this problem I would be grateful if you could share.

Regards

t0mmmy90:
Maybe I didn't understand it correctly but I think that in doc descriptions for bellowed functions should be swapped cause actually calling __nl_cache_pickup() with checkdup set to 1 is causing updates.

```
/**
 * Pickup a netlink dump response and put it into a cache.
 * @arg sk		Netlink socket.
 * @arg cache		Cache to put items into.
 *
 * Waits for netlink messages to arrive, parses them and puts them into
 * the specified cache.
 *
 * @return 0 on success or a negative error code.
 */
int nl_cache_pickup_checkdup(struct nl_sock *sk, struct nl_cache *cache)
{
	return __nl_cache_pickup(sk, cache, 1);
}

/**
 * Pickup a netlink dump response and put it into a cache.
 * @arg sk		Netlink socket.
 * @arg cache		Cache to put items into.
 *
 * Waits for netlink messages to arrive, parses them and puts them into
 * the specified cache. If an old object with same key attributes is
 * present in the cache, it is replaced with the new object.
 * If the old object type supports an update operation, an update is
 * attempted before a replace.
 *
 * @return 0 on success or a negative error code.
 */
int nl_cache_pickup(struct nl_sock *sk, struct nl_cache *cache)
{
	return __nl_cache_pickup(sk, cache, 0);
}
```

PR #277, opened 2021-04-08
lib: check duplicates in cache refill, fix descriptions

t0mmmy90:
Using `nl_cache_pickup_checkdup` while refilling the cache resolves the issue which I reported some time ago #271.

I've also swapped descriptions of `nl_cache_pickup` and `nl_cache_pickup_checkdup ` cause it looks like the second one performs update.

### DIFF
--- a/lib/cache.c
+++ b/lib/cache.c
@@ -751,7 +751,10 @@ static int __nl_cache_pickup(struct nl_sock *sk, struct nl_cache *cache,
  * @arg cache		Cache to put items into.
  *
  * Waits for netlink messages to arrive, parses them and puts them into
- * the specified cache.
+ * the specified cache. If an old object with same key attributes is
+ * present in the cache, it is replaced with the new object.
+ * If the old object type supports an update operation, an update is
+ * attempted before a replace.
  *
  * @return 0 on success or a negative error code.
  */
@@ -766,10 +769,7 @@ int nl_cache_pickup_checkdup(struct nl_sock *sk, struct nl_cache *cache)
  * @arg cache		Cache to put items into.
  *
  * Waits for netlink messages to arrive, parses them and puts them into
- * the specified cache. If an old object with same key attributes is
- * present in the cache, it is replaced with the new object.
- * If the old object type supports an update operation, an update is
- * attempted before a replace.
+ * the specified cache.
  *
  * @return 0 on success or a negative error code.
  */
@@ -1055,7 +1055,7 @@ restart:
 		NL_DBG(2, "Updating cache %p <%s> for family %u, request sent, waiting for reply\n",
 		       cache, nl_cache_name(cache), grp ? grp->ag_family : AF_UNSPEC);
 
-		err = nl_cache_pickup(sk, cache);
+		err = nl_cache_pickup_checkdup(sk, cache);
 		if (err == -NLE_DUMP_INTR) {
 			NL_DBG(2, "Dump interrupted, restarting!\n");
 			goto restart;


thom311:
I took part of the patch, about the fix for the description.


the remainder
```
 
-                 err = nl_cache_pickup(sk, cache);
+                 err = nl_cache_pickup_checkdup(sk, cache);
                  if (err == -NLE_DUMP_INTR) {
```
is probably correct. I didn't apply it (yet), because for most object types it doesn't make sense (because during a dump there should be no duplicated). So all types would pay the overhead of looking first into the cache. That seems dangerous to cause a regression.

I am a bit hesitant to apply the second part, because I think the route cache has other problems as well (e.g. that `ip route replace` replaces a route, and libnl3 doesn't properly handle that; or that the concept of identity is not correct for IPv4 routes (and possibly not for IPv6 routes)).

thom311:
The route handling (when putting routes into the cache) has many issues in libnl3.

For example, a cache requires some sort of identity for its objects. Specifically, if you can add two routes in kernel (using `ip route append`) that only differ by one attribute, then that attribute is part of the identity. For IPv4, NetworkManager does [this](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/c20e3a72e24be065baa69bab3d52f5c1b6f1d0a8/src/libnm-platform/nm-platform.c#L8351) and for IPv6 it does [this](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/c20e3a72e24be065baa69bab3d52f5c1b6f1d0a8/src/libnm-platform/nm-platform.c#L8544). Compare to libnl's identity [here](https://github.com/thom311/libnl/blob/main/lib/route/route_obj.c#L1485). You see that it's wrong. Also NetworkManager's list of what constitutes the identity of a route in kernel is possibly incomplete, at least because kernel coul add a new attribute anytime. The proper solution would be if routes had a unique 64 bit identifier (large enough to never overflow).

Also, the cache relies on events (`RTM_NEWROUTE`, `RTM_DELROUTE`) that tell that an object was removed. But `ip route change` can replace *another* route (as far as the identity is concerned). All it does is send a `RTM_NEWROUTE` message (for the new route) with a `NLM_F_REPLACE` flag. I guess, we then have to search the cache for the route that was replaced... and kernel keeps routes sorted, so we need to find the *first* matching route that was replaced (NetworkManager calls that `NM_PLATFORM_IP_ROUTE_CMP_TYPE_WEAK_ID`). That seems easy to get wrong, and is not very efficient (NetworkManager maintains two indexes for all the routes, so that it can find the replaced route fast -- but maintaining an extra index is overhead too).

IPv6 multihop routes (ECMP) are treated special. On the one hand, the next-hop is part of the identity. On the other hand, when you add a second address that only differs by next-hop, then kernel will merge the existing routes. This is relatively easy to solve, by just pretending that multihop routes don't exist. If you have a multiphop route with `n` next-hops, then it's really `n` single-hop routes. The only twist is, that those routes will be announced in on `RTM_NEWROUTE` message. ... and, that `ip route change` not only replaces one of these routes, but all of them.

TL;DR: it's a mess. But kernel makes it hard to reasonably place routes in a cache.


Of course it should be fixed. But without first adding sensible unit-tests for all of this, I don't feel confident to merge this (and quite possibly, this patch is not sufficient).

I will add some unit-tests first...


Sorry for the disappointing reply.

thom311:
what I commented [here](https://github.com/thom311/libnl/pull/293#issuecomment-1649608175) applies.

I am sorry, but it does not feel right to merge this as-is. See ^^ for why. Closing.

PR #283, opened 2021-05-25
Fix some coverity warnings

bengal:


lib/route/cls/cgroup.c
```diff
@@ -32,15 +32,10 @@ static int cgroup_clone(void *_dst, void *_src)
...
-
-	dst->cg_mask = src->cg_mask;
-	dst->cg_ematch = rtnl_ematch_tree_clone(src->cg_ematch);
-	if (!dst) {
-		free(dst);
-		return -NLE_NOMEM;
+	if (src->cg_ematch) {
+		dst->cg_ematch = rtnl_ematch_tree_clone(src->cg_ematch);
+		if (!dst->cg_ematch)
+			return -NLE_NOMEM;
```
thom311, line 38: this patch is somehow not right. It doesn't use the `_dst` parameter, but also does not initialized `dst` variable.

lib/route/qdisc/netem.c
```diff
@@ -905,10 +907,10 @@ int rtnl_netem_set_delay_distribution(struct rtnl_qdisc *qdisc, const char *dist
 	int n = 0;
 	size_t i;
 	size_t len = 2048;
-	char *line;
+	_nl_auto_free char *line = NULL;
 	char name[NAME_MAX];
 	char dist_suffix[] = ".dist";
-	int16_t *data;
+	_nl_auto_free int16_t *data = NULL;
 	char *test_suffix;
```
thom311, line 914: the patch is right, but could you please do a followup to handle `ENOMEM` from `calloc()` calls?

thom311 (changes requested):

lib/route/cls/cgroup.c
bengal: Fixed.

lib/route/qdisc/netem.c
bengal: Done.

PR #290, opened 2021-07-22
route: check if nh exists while updating route

t0mmmy90:


thom311:
I think, how libnl3 handles routes (and IPv6 multipath routes specifically) has several issues.

Still, the patch makes sense -- in the context of what libnl3 currently does.

Applied. I did however add a followup-fix. Please review.


thank you!!

PR #292, opened 2021-08-04
route/link: check calloc() return value

bengal:


Karry:
Even when documentation is saying that `calloc` may return NULL on Out of memory, I am almost sure that it will never happen on real systems. Linux kernel more likely invokes OOM killer. Possibly some obscure embedded systems / libc?

thom311:
in general, libnl3 should try to handle all ENOMEM conditions correctly. Patches welcome.

I think there are however various subtle places where this isn't done correctly.

Merged. Thanks Beniamino!!

Issue #294, opened 2021-09-10
Compile error with bison 3.8 - yyerror - conflicting types - static declaration follows non-static declaration - POSIX Yacc does not support


heitbaum:
Successful compile with GNU Bison 3.7.6 and errors with Bison 3.8

Reference from bison mailing list: https://www.mail-archive.com/bison-patches@gnu.org/msg07739.html

Full Logs:
- http://paste.ubuntu.com/p/yT9XNPNfs3/

### POSIX Yacc - warning
```
lib/route/cls/ematch_syntax.y:26.1-14: warning: POSIX Yacc does not support %error-verbose [-Wyacc]
   26 | %error-verbose
      | ^~~~~~~~~~~~~~
lib/route/cls/ematch_syntax.y:26.1-14: warning: deprecated directive: '%error-verbose', use '%define parse.error verbose' [-Wdeprecated]
   26 | %error-verbose
      | ^~~~~~~~~~~~~~
      | %define parse.error verbose
lib/route/cls/ematch_syntax.y:27.1-7: warning: POSIX Yacc does not support %define [-Wyacc]
   27 | %define api.pure
      | ^~~~~~~
lib/route/cls/ematch_syntax.y:28.14-22: warning: POSIX Yacc does not support string literals [-Wyacc]
   28 | %name-prefix "ematch_"
      |              ^~~~~~~~~
lib/route/cls/ematch_syntax.y:28.1-22: warning: POSIX Yacc does not support %name-prefix  [-Wyacc]
   28 | %name-prefix "ematch_"
      | ^~~~~~~~~~~~~~~~~~~~~~
lib/route/cls/ematch_syntax.y:28.1-22: warning: deprecated directive: '%name-prefix "ematch_"', use '%define api.prefix {ematch_}' [-Wdeprecated]
   28 | %name-prefix "ematch_"
      | ^~~~~~~~~~~~~~~~~~~~~~
      | %define api.prefix {ematch_}
lib/route/cls/ematch_syntax.y:59.20-22: warning: POSIX Yacc does not support string literals [-Wyacc]
   59 | %token <i> KW_OPEN "("
      |                    ^~~
lib/route/cls/ematch_syntax.y:60.21-23: warning: POSIX Yacc does not support string literals [-Wyacc]
   60 | %token <i> KW_CLOSE ")"
      |                     ^~~
lib/route/cls/ematch_syntax.y:61.20-22: warning: POSIX Yacc does not support string literals [-Wyacc]
[...]

PR #295, opened 2021-09-10
yyerror: update to POSIX standard

heitbaum:
To comply with the latest POSIX standard, in Yacc compatibility mode
(options `-y`/`--yacc`) Bison now generates prototypes for yyerror and
yylex.  In some situations, this is breaking compatibility: if the user
has already declared these functions but with some differences (e.g., to
declare them as static, or to use specific attributes), the generated
parser will fail to compile.  To disable these prototypes, #define yyerror
(to `yyerror`), and likewise for yylex.

refer: https://git.savannah.gnu.org/cgit/bison.git/tree/NEWS

GNU Bison 3.8

Issue #294 has the details of the errors

### DIFF
--- a/lib/route/cls/ematch_syntax.y
+++ b/lib/route/cls/ematch_syntax.y
@@ -46,6 +46,7 @@
 %{
 extern int ematch_lex(YYSTYPE *, void *);
 
+#define ematch_error yyerror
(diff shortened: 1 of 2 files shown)


thom311:
merged. Thank you!!

PR #296, opened 2021-09-10
xfrm: Add support for xfrm user offloading

spellingmistake:


thom311:
merged to master.

I did two follow-up patches, that fix something what I think was wrong. If you disagree, please report.

Thanks for the patch!!

spellingmistake:
Hi Thomas,
sorry for the late reply. Your correction is totally valid. Your github message however went unnoticed (as well as the bug in the code). I only used the code to **add** states with the offloading flag. Parsing was only an addon to complete the patch.
Thanks for merging and the fix
Thomas

PR #300, opened 2021-11-29
Supporting Hardware offload capability for MACsec

johandroid:


thom311:
merged to master. Thanks Johan!!

PR #305, opened 2022-03-30
route/mdb: fixes to dumping, support for multicast mac entries, minor fixes to nl-monitor

troglobit:
This is a mixed bag of fixes related to debugging the bridge mdb support.

 1. Minor fixes and additions to `nl-monitor`, used to debug mdb system
 2. Add missing details mdb dump callback 
 3. Add support for MAC multicast entries

With this series applied, on a system with a bridge set up with at least one port:

    # nl-monitor mdb &
    # bridge mdb add dev br0 port eth0 grp 01:02:03:c0:ff:ee vid 1 permanent
    dev 9
    port 3 vid 1 proto 0x0000 address 01:02:03:c0:ff:ee

### DIFF
--- a/include/linux-private/linux/if_bridge.h
+++ b/include/linux-private/linux/if_bridge.h
@@ -243,6 +243,7 @@ struct br_mdb_entry {
 		union {
 			__be32	ip4;
 			struct in6_addr ip6;
+			unsigned char mac_addr[ETH_ALEN];
 		} u;
 		__be16		proto;
 	} addr;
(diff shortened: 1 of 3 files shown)


thom311:
merged.

thank you!!


Sidenote: when importing linux headers to `include/linux-private`, maybe we should always take the entire file from a recent kernel release. Picking only parts of it, results in the file being a compose of different versions. That seems error prone. Granted, if we reimport only one file at a time, then they are picked from different versions anyway. Maybe there should be a script to update all linux headers at once to a new version. Dunno. Anyway, next time please do one separate commit that completely updates the linux header as a whole (one file at a time, unless we create a script to re-import them all).

troglobit:
@thom311 sure, will do!  I thought about doing that at first, and you're right, that's a much better approach.

(Continuing work on adding support for add/del MDB entries next week :)

PR #309, opened 2022-04-19
cls: flower: extend flower API

voldymyr:
The following API has been added:

    rtnl_flower_set_ipv4_src
    rtnl_flower_get_ipv4_src
    rtnl_flower_set_ipv4_dst
    rtnl_flower_get_ipv4_dst

### DIFF
--- a/lib/route/cls/flower.c
+++ b/lib/route/cls/flower.c
@@ -27,6 +27,10 @@
 #define FLOWER_ATTR_IP_DSCP       (1 << 9)
 #define FLOWER_ATTR_IP_DSCP_MASK  (1 << 10)
 #define FLOWER_ATTR_PROTO         (1 << 11)
+#define FLOWER_ATTR_IPV4_SRC      (1 << 12)
+#define FLOWER_ATTR_IPV4_SRC_MASK (1 << 13)
+#define FLOWER_ATTR_IPV4_DST      (1 << 14)
+#define FLOWER_ATTR_IPV4_DST_MASK (1 << 15)
 /** @endcond */
 
 #define FLOWER_DSCP_MAX             0xe0
@@ -46,6 +50,10 @@ static struct nla_policy flower_policy[TCA_FLOWER_MAX + 1] = {
 	[TCA_FLOWER_KEY_IP_TOS]        = { .type = NLA_U8 },
 	[TCA_FLOWER_KEY_IP_TOS_MASK]   = { .type = NLA_U8 },
 	[TCA_FLOWER_KEY_VLAN_ETH_TYPE] = { .type = NLA_U16 },
+	[TCA_FLOWER_KEY_IPV4_SRC]      = { .type = NLA_U32 },
+	[TCA_FLOWER_KEY_IPV4_SRC_MASK] = { .type = NLA_U32 },
+	[TCA_FLOWER_KEY_IPV4_DST]      = { .type = NLA_U32 },
+	[TCA_FLOWER_KEY_IPV4_DST_MASK] = { .type = NLA_U32 },
 };
 
 static int flower_msg_parser(struct rtnl_tc *tc, void *data)
@@ -121,6 +129,26 @@ static int flower_msg_parser(struct rtnl_tc *tc, void *data)
 		f->cf_mask |= FLOWER_ATTR_IP_DSCP_MASK;
 	}
 
+	if (tb[TCA_FLOWER_KEY_IPV4_SRC]) {
+		f->cf_ipv4_src = nla_get_u32(tb[TCA_FLOWER_KEY_IPV4_SRC]);
+		f->cf_mask |= FLOWER_ATTR_IPV4_SRC;
+	}
+
+	if (tb[TCA_FLOWER_KEY_IPV4_SRC_MASK]) {
+		f->cf_ipv4_src_mask = nla_get_u32(tb[TCA_FLOWER_KEY_IPV4_SRC_MASK]);
+		f->cf_mask |= FLOWER_ATTR_IPV4_SRC_MASK;
+	}
+
+	if (tb[TCA_FLOWER_KEY_IPV4_DST]) {
+		f->cf_ipv4_dst = nla_get_u32(tb[TCA_FLOWER_KEY_IPV4_DST]);
+		f->cf_mask |= FLOWER_ATTR_IPV4_DST;
+	}
+
+	if (tb[TCA_FLOWER_KEY_IPV4_DST_MASK]) {
+		f->cf_ipv4_dst_mask = nla_get_u32(tb[TCA_FLOWER_KEY_IPV4_DST_MASK]);
+		f->cf_mask |= FLOWER_ATTR_IPV4_DST_MASK;
+	}
+
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


lib/route/cls/flower.c
```diff
@@ -598,6 +670,124 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+}
+
+/**
+ * Set IPv4 destination address for flower classifier
+ * @arg cls		Flower classifier.
+ * @arg addr		IPv4 destination address
+ * @arg mask		mask for IPv4 destination address
+ * @return 0 on success or a negative error code.
+ */
+int rtnl_flower_set_ipv4_dst(struct rtnl_cls *cls, int addr, int mask)
```
thom311, line 739: `int` is not guaranteed to be a 32 bit integer (as far as C is concerned... though, I think on Linux we are guaranteed that `int` is at least 32 bits large).

also, it seems a bit ugly to use a signed integer for an IPv4 address.

We could use `uint32_t`, but I think the right choice is `in_addr_t`. The only downside of `in_addr_t`, is that we need to include `<netinet/in.h>`. Interestingly, we already use `struct in_addr` in `include/netlink/route/cls/u32.h` (is that a bug?). It seems to me that `<netinet/in.h>` is such a fundamental dependency, that we can afford to include it in libnl3's public headers (and thus have `in_addr_t`).

lib/route/cls/flower.c
```diff
@@ -598,6 +670,124 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+
+/**
+ * Get IPv4 destination address for flower classifier
+ * @arg cls		Flower classifier.
+ * @arg addr		IPv4 destination address
+ * @arg mask		mask for IPv4 destination address
+ * @return 0 on success or a negative error code.
+ */
+int rtnl_flower_get_ipv4_dst(struct rtnl_cls *cls, struct nl_addr *addr,
+			     uint32_t *mask)
```
thom311, line 773: it would seem that the IPv4 is always a `uint32_t` (actually `in_addr_t` or `struct in_addr`, see above).

This is not something that could change, and the function name (`rtnl_flower_get_ipv4_dst()`) also ties that down. I think it's needlessly complicated to expose the address as `struct nl_addr`, in particular, since `mask` is already a `uint32_t`. I think the function should be
```C
int rtnl_flower_get_ipv4_dst(struct rtnl_cls *cls, in_addr_t *out_addr, in_addr_t *out_mask);
```

lib/route/cls/flower.c
```diff
@@ -598,6 +670,124 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+
+	if (addr) {
+		f->cf_ipv4_src = nl_addr_build(AF_INET,
+					       (unsigned char *)&addr,
+					       sizeof(__be32));
+
+		f->cf_mask |= FLOWER_ATTR_IPV4_SRC;
+
+		if (mask) {
+			uint32_t m = 0xFFFFFFFF << (32 - mask);
```
thom311, line 695: it seems that `TCA_FLOWER_KEY_IPV4_DST_MASK` is a 32bit mask (effectively type `in_addr_t`), and not a prefix length (from 0 to 32).

lib/route/cls/flower.c
```diff
@@ -598,6 +670,124 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+	if (!(f = rtnl_tc_data_peek(TC_CAST(cls))))
+		return -NLE_INVAL;
+
+	if (!(f->cf_mask & FLOWER_ATTR_IPV4_DST))
+		return -NLE_MISSING_ATTR;
+
+	if (addr)
+		memcpy(addr, f->cf_ipv4_dst, AF_INET);
+
+	*mask = f->cf_ipv4_dst_mask;
```
thom311, line 786: there is one getter for both the address and the mask (which is good).
But internally, you track this in two separate fields, which are even announced on netlink separately.

It's not clear, why we check above for `if (!(f->cf_mask & FLOWER_ATTR_IPV4_DST))`, but here we access the mask even if it's not set. It seems, if the mask is not set, then it should default to 0xFFFFFFFF. Currently it defaults to zero.

lib/route/cls/flower.c
```diff
@@ -598,6 +670,124 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+	struct rtnl_flower *f;
+
+	if (!(f = rtnl_tc_data_peek(TC_CAST(cls))))
+		return -NLE_INVAL;
+
+	if (!(f->cf_mask & FLOWER_ATTR_IPV4_DST))
+		return -NLE_MISSING_ATTR;
+
+	if (addr)
+		memcpy(addr, f->cf_ipv4_dst, AF_INET);
```
thom311, line 784: you cannot memcpy as `struct nl_addr` this way. anyway, I think this should be a `in_addr_t`, then you can just copy it.

thom311 (commented):
thanks for the patch!!

thom311:
Seems there is work needed on this PR. I am going to close this.
But please reopen if you'd like to move this forward!
Thank you.

voldymyr:
Hello,

I've recently fixed the code accordingly and want to re-open this issue. Please have a look again.

Thanks!

thom311:
please push latest version of the branch. Possibly rebase first. Thank you.

voldymyr:
Rebased and pushed.

lib/route/cls/flower.c
```diff
@@ -598,6 +649,126 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+ */
+int rtnl_flower_set_ipv4_src(struct rtnl_cls *cls, in_addr_t addr, in_addr_t mask)
+{
+	struct rtnl_flower *f;
+
+	if (!(f = rtnl_tc_data(TC_CAST(cls))))
+		return -NLE_NOMEM;
+
+	if (addr) {
+		f->cf_ipv4_src = htonl(addr);
```
thom311, line 667: I think all those `htonl` are wrong. You copy from an `in_addr_t` to an `in_addr_t`. Neither of them is native endianness.

lib/route/cls/flower.c
```diff
@@ -272,6 +313,16 @@ static void flower_dump_details(struct rtnl_tc *tc, void *data,
 
 	if (f->cf_mask & FLOWER_ATTR_IP_DSCP_MASK)
 		nl_dump(p, " dscp_mask %u", f->cf_ip_dscp_mask);
+
+	if (f->cf_mask & FLOWER_ATTR_IPV4_SRC) {
+		inet_ntop(AF_INET, &f->cf_ipv4_src, addr_str, sizeof(addr_str));
+		nl_dump(p, "IPv4 src %s mask %u\n", addr_str, ntohl(f->cf_ipv4_src_mask));
```
thom311, line 319: the mask is also just an IP address. Print via `inet_ntop()` too.

lib/route/cls/flower.c
```diff
@@ -598,6 +649,126 @@ int rtnl_flower_get_ip_dscp(struct rtnl_cls *cls, uint8_t *dscp, uint8_t *mask)
...
+		f->cf_ipv4_dst = htonl(addr);
+		f->cf_mask |= FLOWER_ATTR_IPV4_DST;
+
+		if (mask) {
+			f->cf_ipv4_dst_mask = htonl(mask);
+			f->cf_mask |= FLOWER_ATTR_IPV4_DST_MASK;
+		}
+
+		return 0;
+	}
```
thom311, line 736: this does not seem right. The current code does:

- fail if `addr==0` and do nothing. Why?
- only touch mask if `mask!=0`. Why?

I think the setters should not go overboard with validity checking. Especially if it's done inconsistently.

- Is `mask==0` something disallowed by kernel? If so, we could use that to unset `FLOWER_ATTR_IPV4_DST_MASK`.
- Is `addr==0` something disallowed by kernel? If so, we could use that to unset `FLOWER_ATTR_IPV4_DST`.

Otherwise, the setter cannot express to unset `cf_mask`, and still support setting any value, uncluding `0.0.0.0`. Not being able to unset it is not nice (if having it unset is a possible configuration that can be achieved otherwise).

lib/route/cls/flower.c
voldymyr: Since we've spoken about this matter in other PR, there is no need to bring this here again. I will adapt functions to expect _address/mask_ in network byte order.

thom311:
merged. Thanks @voldymyr !!

PR #311, opened 2022-04-25
route/link: Remove reset of cache ops when cloning a link

jackuess:
We are running in to problems with working with a cloned link cache. We are more or less doing:

```
rtnl_link_alloc_cache(sk, AF_BRIDGE, &cache);
new = nl_cache_clone(cache);

link = ...some link fetched from new
rtnl_link_is_bridge(link);  /* Will always return 0 since link_clone has set link->l_af_ops = NULL */
```

As mentioned `rtnl_link_is_bridge` will return 0 even for links that are bridges, since link_clone remove the cache_ops when the cache was cloned.

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -334,6 +334,9 @@ static int link_clone(struct nl_object *_dst, struct nl_object *_src)
 	if ((err = do_foreach_af(src, af_clone, dst)) < 0)
 		return err;
 
+	if (src->l_af_ops)
+		dst->l_af_ops = af_lookup_and_alloc(dst, src->l_af_ops->ao_family);
+
 	if (src->l_phys_port_id)
 		if (!(dst->l_phys_port_id = nl_data_clone(src->l_phys_port_id)))
 			return -NLE_NOMEM;


thom311:
I think this is not the right fix. At the very least, this does not get the ref-counting for the `l_af_ops` pointer right. It also does not take proper care of the associated data.

At the start of `link_clone()`, `dst` was initialized with `memcpy()` from `src`. So all the pointers there are aliased/owned by two objects. As first thing, the function clears all pointer values, with the intent to set them afterwards again (this time by fixing the ref-count or clone the data).

Below there is
```C
    if ((err = do_foreach_af(src, af_clone, dst)) < 0)
        return err;
```
why does that not properly clone everything?

troglobit:
@thom311 That's a very good question.  We tried to figure out if this the `l_af_ops` used refcounters, and if something fancier was needed, but all subsystems we could see that employed the `l_af_ops` was only using callbacks.  We can continue looking at this today and see if we can find something more substantial.  (We think bridge, inet, and inet6 are affected by this.)

jackuess:
Thanks @thom311, I definitely overlooked the ref-counting.

I'm not sure I understand how `af_clone` could clone the ops struct, but I have now force pushed another approach, that is hopefully better than my initial, that uses af_lookup_and_alloc - which should increase the ref-count properly.

thom311:
I also find the entire "ops" thing non-obvious. The last version lgtm.

merged. Thank you!!

Issue #308, opened 2022-04-16
Test failure with 3.6.0 (nl-test-util.c:69:S:Core:cache_and_clone:0: Assertion 'f' failed)


thesamesam:
Noticed when packaging 3.6.0 for Gentoo. Not hitting #306 anymore as expected, but a new one!

test-suite.log:
```
===================================
   libnl 3.6.0: ./test-suite.log
===================================

# TOTAL: 1
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: tests/check-all
=====================

Running suite(s): main
 Abstract addresses
 Netlink attributes
 Clone ematch tree
 netns
90%: Checks: 11, Failures: 1, Errors: 0
/var/tmp/portage/dev-libs/libnl-3.6.0/work/libnl-3.6.0/tests/nl-test-util.c:69:S:Core:cache_and_clone:0: Assertion 'f' failed
FAIL tests/check-all (exit status: 1)
```

System information:
```
Portage 3.0.30 (python 3.10.4-final-0, default/linux/amd64/17.1/hardened, gcc-11.2.1, glibc-2.35-r2, 5.15.34-gentoo-dist-hardened x86_64)
=================================================================
System uname: Linux-5.15.34-gentoo-dist-hardened-x86_64-AMD_Ryzen_9_3950X_16-Core_Processor-with-glibc2.35
KiB Mem:    65830320 total,  21308240 free
KiB Swap:   25067512 total,  25067512 free
Timestamp of repository gentoo: Sat, 16 Apr 2022 00:35:04 +0000
Head commit of repository gentoo: ac450fcad2bd337614917dd0edcfe6f0110c7a9c

Timestamp of repository kde: Fri, 15 Apr 2022 22:49:46 +0000
Head commit of repository kde: 23c4d80d4dd6a254cac108605182e9bc63d2646d

Timestamp of repository qt: Fri, 15 Apr 2022 12:19:40 +0000
Head commit of repository qt: 94bd1952996bade333941be20dca33b60c3f48a8

Timestamp of repository sam_c: Fri, 15 Apr 2022 12:19:54 +0000
Head commit of repository sam_c: dce8902550901f857f19d0e43f01c4fde8173605

Timestamp of repository steam-overlay: Fri, 15 Apr 2022 12:19:42 +0000
[...]

PR #314, opened 2022-05-08
[th/test-link] add unit tests creating some interfaces

thom311:
this is taken from https://src.fedoraproject.org/rpms/libnl3/pull-request/4


thom311:
@thesamesam . Hi. I merged here a new unit test to `main`. It passes on github-actions and also when building a Fedora package (which uses a `mock` environment). I'd be curious if it can pass in the environment in which you ran #308... Otherwise, we might notice only after release. Anyway, if not possible, it's fine. Thanks.

PR #316, opened 2022-05-19
flower: use correct attribute when filling out flags

voldymyr:


thom311:
merged. Thank you!!

PR #317, opened 2022-05-23
route/act: add NAT action

voldymyr:
Signed-off-by: Magnus Öberg <magnus.oberg@westermo.se>
Signed-off-by: Volodymyr Bendiuga <volodymyr.bendiuga@gmail.com>

### DIFF
--- a/include/netlink/route/act/nat.h
+++ b/include/netlink/route/act/nat.h
@@ -1,0 +1,35 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/*
+ * Copyright (c) 2016 Magnus Öberg <magnus.oberg@westermo.se>
+ */
+
+#ifndef NETLINK_NAT_H_
+#define NETLINK_NAT_H_
+
+#include <netlink/netlink.h>
+#include <netlink/cache.h>
+#include <netlink/route/action.h>
+#include <linux/tc_act/tc_nat.h>
+
+#include <netinet/in.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+extern int rtnl_nat_set_old_addr(struct rtnl_act *act, in_addr_t addr);
+extern int rtnl_nat_get_old_addr(struct rtnl_act *act, in_addr_t *addr);
+extern int rtnl_nat_set_new_addr(struct rtnl_act *act, in_addr_t addr);
+extern int rtnl_nat_get_new_addr(struct rtnl_act *act, in_addr_t *addr);
+extern int rtnl_nat_set_mask(struct rtnl_act *act, in_addr_t bitmask);
+extern int rtnl_nat_get_mask(struct rtnl_act *act, in_addr_t *bitmask);
+extern int rtnl_nat_set_flags(struct rtnl_act *act, uint32_t flags);
+extern int rtnl_nat_get_flags(struct rtnl_act *act, uint32_t *flags);
+extern int rtnl_nat_set_action(struct rtnl_act *act, int action);
+extern int rtnl_nat_get_action(struct rtnl_act *act, int *action);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif	/* NETLINK_NAT_H */
--- a/lib/route/act/nat.c
+++ b/lib/route/act/nat.c
@@ -1,0 +1,288 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/*
+ * Copyright (c) 2016 Magnus Öberg <magnus.oberg@westermo.se>
+ */
+
+/**
+ * @ingroup act
+ * @defgroup act_nat NAT
+ *
+ * @{
+ */
+
+#include <netlink-private/netlink.h>
+#include <netlink-private/tc.h>
+#include <netlink/netlink.h>
+#include <netlink/attr.h>
+#include <netlink/utils.h>
+#include <netlink-private/route/tc-api.h>
+#include <netlink/route/act/nat.h>
+#include <netlink/route/tc.h>
+
+static struct nla_policy nat_policy[TCA_NAT_MAX + 1] = {
+	[TCA_NAT_PARMS] = { .minlen = sizeof(struct tc_nat) },
+};
+
+/**
+ * nat operations
+ */
+
+static int nat_msg_parser(struct rtnl_tc *tc, void *data)
+{
+	struct tc_nat *nat = data;
+	struct nlattr *tb[TCA_NAT_MAX + 1];
+	int err;
+
+	err = tca_parse(tb, TCA_NAT_MAX, tc, nat_policy);
+	if (err < 0)
+		return err;
+
+	if (!tb[TCA_NAT_PARMS])
+		return -NLE_MISSING_ATTR;
+
+	nla_memcpy(nat, tb[TCA_NAT_PARMS], sizeof(*nat));
+
+	return NLE_SUCCESS;
+}
+
+static void nat_free_data(struct rtnl_tc *tc, void *data)
+{
+}
+
+static int nat_msg_fill(struct rtnl_tc *tc, void *data, struct nl_msg *msg)
+{
+	struct tc_nat *nat = data;
+
+	if (!nat)
+		return -NLE_OBJ_NOTFOUND;
+
+	NLA_PUT(msg, TCA_NAT_PARMS, sizeof(*nat), nat);
+
+	return NLE_SUCCESS;
+
+nla_put_failure:
+	return -NLE_NOMEM;
+}
+
+static void nat_dump_line(struct rtnl_tc *tc, void *data,
+                          struct nl_dump_params *p)
+{
+	struct tc_nat *nat = data;
+	char buf[32];
+	uint32_t mask;
+	int pfx = 0;
+
+	if (!nat)
+		return;
+
+	if (nat->flags & TCA_NAT_FLAG_EGRESS)
+		nl_dump(p, " egress");
+	else
+		nl_dump(p, " ingress");
+
+	mask = nat->mask;
+	while (mask > 0) {
+		mask = mask >> 1;
+		pfx++;
+	}
+
+	inet_ntop(AF_INET, &nat->old_addr, buf, sizeof(buf));
+	nl_dump(p, " %s", buf);
+	if (pfx < 32)
+		nl_dump(p, "/%d", pfx);
+
+	inet_ntop(AF_INET, &nat->new_addr, buf, sizeof(buf));
+	nl_dump(p, " %s", buf);
+	if (pfx < 32)
+		nl_dump(p, "/%d", pfx);
+}
+
+/**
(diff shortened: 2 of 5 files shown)


libnl-route-3.sym
```diff
@@ -1250,3 +1250,17 @@ global:
...
+	rtnl_nat_set_old_addr;
+	rtnl_nat_get_old_addr;
+	rtnl_nat_set_new_addr;
+	rtnl_nat_get_new_addr;
+	rtnl_nat_set_mask;
+	rtnl_nat_get_mask;
+	rtnl_nat_set_flags;
+	rtnl_nat_get_flags;
+	rtnl_nat_set_action;
+	rtnl_nat_get_action;
```
thom311, line 1265: please sort symbols by name. `vim` can do that nicely ( https://vim.fandom.com/wiki/Sort_lines )

include/netlink/route/act/nat.h
```diff
@@ -0,0 +1,41 @@
...
+extern int rtnl_nat_set_flags(struct rtnl_act *act, uint32_t flags);
+extern int rtnl_nat_get_flags(struct rtnl_act *act, uint32_t *flags);
+extern int rtnl_nat_set_action(struct rtnl_act *act, int action);
+extern int rtnl_nat_get_action(struct rtnl_act *act, int *action);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif	/* NETLINK_NAT_H */
```
thom311, line 41: I will reformat all new files with clang-format. This only affects new code, and it does not affect files under `include/linux-private/`. There is a `.clang-format` file for the correct style. The problem is, that we will need to agree on a version of clang-format (because in subtle cases, the clang-format version causes different formatting). So if you are willing, please reformat the files. Or don't bother, and I'll do it before merge.

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+	return NLE_SUCCESS;
+}
+
+int rtnl_nat_get_action(struct rtnl_act *act, int *action)
+{
+        struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data_peek(TC_CAST(act))))
+	        return -NLE_NOMEM;
```
thom311, line 267: nomem is not right. Maybe `-NLE_NOATTR`?

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+	return NLE_SUCCESS;
+}
+
+int rtnl_nat_get_flags(struct rtnl_act *act, uint32_t *flags)
+{
+	struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data(TC_CAST(act))))
+		return -NLE_NOMEM;
```
thom311, line 240: `rtnl_tc_data_peek()` and `NLE_NOATTR` ?

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+	return NLE_SUCCESS;
+}
+
+int rtnl_nat_get_mask(struct rtnl_act *act, in_addr_t *bitmask)
+{
+	struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data(TC_CAST(act))))
+		return -NLE_NOMEM;
```
thom311, line 206: `rtnl_tc_data_peek()` and `NLE_NOATTR` ?

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+	return NLE_SUCCESS;
+}
+
+int rtnl_nat_get_old_addr(struct rtnl_act *act, in_addr_t *addr)
+{
+	struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data(TC_CAST(act))))
+		return -NLE_NOMEM;
```
thom311, line 144: `rtnl_tc_data_peek()` and `NLE_NOATTR` ?

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+	return NLE_SUCCESS;
+}
+
+int rtnl_nat_get_new_addr(struct rtnl_act *act, in_addr_t *addr)
+{
+	struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data(TC_CAST(act))))
+		return -NLE_NOMEM;
```
thom311, line 175: `rtnl_tc_data_peek()` and `NLE_NOATTR` ?

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+
+static void nat_free_data(struct rtnl_tc *tc, void *data)
+{
+}
+
+static int nat_clone(void *_dst, void *_src)
+{
+	struct tc_nat *dst = _dst, *src = _src;
+
+	memcpy(dst, src, sizeof(*src));
```
thom311, line 62: this clone function only does a plain memcpy. I don't think you need to implement it, the default implementation should do this already. Please check whether this function can be dropped.

lib/route/act/nat.c
```diff
@@ -0,0 +1,303 @@
...
+ * @return 0 on success or negative error code in case of an error.
+ */
+int rtnl_nat_set_old_addr(struct rtnl_act *act, in_addr_t addr)
+{
+	struct tc_nat *nat;
+
+	if (!(nat = (struct tc_nat *) rtnl_tc_data(TC_CAST(act))))
+		return -NLE_NOMEM;
+
+	nat->old_addr = htonl(addr);
```
thom311, line 134: this `htonl()` is wrong. In `struct tc_nat`, the addresses are already marked as type of `__be32`, that is big-endian == network-byte-order == what also `in_addr_t` is.

You accept here an IP address parameter `in_addr_t addr`, which is commonly understood to be in network byte order already. Also, in `nat_dump_line()` you directly pass `nat->old_addr` ot `inet_ntop()`, indicating that those numbers are also in network-byte order already.

All these `htonl()` must be dropped.

Did you test this? Are you working on a big-endian machine, so you didn't notice?

voldymyr: It is convenient to pass _IPv4 address_ and _mask_ in host byte order and disregard endianness of a system, and let _libnl_ handle this matter internally. This is the primary reason for all _htonl()/ntohl()_ calls.

It is true that _old_addr_ and _new_addr_ are ___be32_, but _in_addr_t_ is of type _uint32_t_ and the endianness here we could decide for ourselves, but it seems that if we prioritize usability of _libnl_, then passing _address/mask_ in host byte order is to be preferred.

I do have access to both big and little endian machines and this code has been tested there numerous times, on different kernel versions including the latest stable (5.17.9).

If you do prefer to have _address/mask_ in network byte order instead then let me know and I will rework that part.

I agree entirely to the rest of your comments and fixes are on the way.
Thanks!

thom311: > but in_addr_t is of type uint32_t

both `__be32` and `in_addr_t` are just typedefs for `uint32_t` -- which in turn is (usually) just a typedef for `unsigned`. So it's true that these types are the same as far as C is concerned.

however, the meaning is different. If you have a variable of type `__be32`, then one very much expects that it contains the number in big-endian, while `uint32_t` is expected to be native endianness.

And `in_addr_t` is always network-byte-order (that is: big endian).

You cannot mix endianness, and when copying from one endianness to another, you need to use conversion functions (like, `htonl()` for host/native to network/big endian).

However, `in_addr_t` and `be32` are the same endianness. If you then use `htonl()` you are wrongly swapping bytes (on little endian machines).

> It is convenient to pass IPv4 address and mask in host byte order and disregard endianness of a system,

This is wrong. It is convenient to track IPv4 addresses in the intended type, that is `in_addr_t`, which is network-byte order and what system function like `inet_ntop()`, `inet_pton()` expect. They do so on little-endian and big-endian machines. `man inet_ntop()` says: "src  points  to  a  struct in_addr (in network byte order)...".

> If you do prefer to have address/mask in network byte order

There is nothing to rework, except dropping the wrong bytes swaps `htonl()`.

If your application parses a string "192.168.0.1" via `inet_pton()`, then the result will be a `in_addr_t` value that already has the right endianness. You need to be clear what endianness your variable has, and in the best case, you keep it network-byte-order all way around (that is: `in_addr_t`).

You say you tested it on little+big endian machines. Then you are doing something odd, possibly needlessly. Anyway libnl3 API should expect `in_addr_t` arguments to be in big-endian.

voldymyr: > > but in_addr_t is of type uint32_t
> 
> both `__be32` and `in_addr_t` are just typedefs for `uint32_t` -- which in turn is (usually) just a typedef for `unsigned`. So it's true that these types are the same as far as C is concerned.
> 
> however, the meaning is different. If you have a variable of type `__be32`, then one very much expects that it contains the number in big-endian, while `uint32_t` is expected to be native endianness.
> 

Yes, I totally agree on the above statements.

> And `in_addr_t` is always network-byte-order (that is: big endian).
> 
The `in_addr_t` type might be slightly confusing though, since it actually can be of any endianness, as is suggested by the common API. Here is the link to some examples: https://linux.die.net/man/3/inet_network
We may find that `in_addr_t inet_addr(const char *cp);` returns an address in network byte order, as the description states. However we do see other functions such as `in_addr_t inet_lnaof(struct in_addr in);` and `in_addr_t inet_netof(struct in_addr in);` returning values in host byte order, according to descriptions.

From this follows that `__be32` type basically states explicitly that the value contained is in big endian format, while `in_addr_t` may actually be of any endianness.

If you still wish _libnl_ to expect values in network byte order then I will remove all the `htonl()/ntohl()` calls.
Let me know what you think. Thanks!

thom311: `in_addr_t inet_netof(struct in_addr in);` returns an `in_addr_t` but is documened that it's host order. Note that these functions are deprecated. Can you find a non-deprecated function that uses `in_addr_t` in host byte order?

OK, so we have posix/libc functions that accept `in_addr_t` and explicitly document in which order the IP address is. I am very sure, we should define that our functions expect `in_addr_t` parameters to be in network-byte order, and not in host-byte order. In particular, because that is also then compatible with `nl_addr_set_binary_addr()` and something you could pass to `inet_ntop()`.

thom311: https://www.gnu.org/software/libc/manual/html_node/Host-Address-Functions.html

>  They represent Internet addresses in network byte order, and network numbers and local-address-within-network numbers in host byte order. 

(you are right, that `in_addr_t` is sometimes used in host order)

voldymyr: Alright, I will adapt functions to expect _address/mask_ in network byte order.
Thanks!

voldymyr:
Updated, rebased and force-pushed. Please have a look again. Thanks.

thom311:
Merged. Thanks!!

PR #319, opened 2022-05-25
route/act: allow full set of actions on gact,skbedit,mirred

voldymyr:


thom311:
This patch goes to the right direction. There is no need for the setter to do such error checking. The user anyway needs to take care to configure correct values, and whether a value is correct does not only depend on the hard-coded integer range but on things that the simple setter couldn't even handling. So, there is little point in rejecting some configuration, while other (invalid) configurations cannot be prevented. Worse, hardcoding the range means that libnl needs patches to support newer kernel features.

But there is no need to even check for `if ((action < TC_ACT_UNSPEC) || (action > TC_ACT_VALUE_MAX))`. We set internally an "int", just allow any value and let kernel tell us whether the value is correct.

voldymyr:
Please have a look, I've made changes and force-pushed. Thanks!

thom311:
merged. Thank you for the patch!!

PR #321, opened 2022-05-31
lib: object: Rectify attribute mask logic in 'nl_object_compare()'

voldymyr:
Setting of the attribute mask could be narrowed down to two cases:

	1. Object possesses either or both 'oo_id_attrs_get' and
	'oo_id_attrs'
	2. Object lacks both 'oo_id_attrs_get' and 'oo_id_attrs'

In the primary case, one would expect to set, for upcoming comparison,
objects 'ce_mask', to already pre-set required attributes. This would
enable comparison of required attributes together with those set in
the object's 'ce_mask', if such exist.

In the secondary case, with no required attributes in place and 64bit
mask used instead, one would expect to set object's 'ce_mask' and
exclude all other bits. This would enable comparison of only those
attributes defined in the object's 'ce_mask'.

This patch takes care of the above mentioned cases, while previous
behavior would neglect the primary case and never actually compare
any attributes set in the object's 'ce_mask', except the required.

### DIFF
--- a/lib/object.c
+++ b/lib/object.c
@@ -338,8 +338,13 @@ int nl_object_identical(struct nl_object *a, struct nl_object *b)
 		req_attrs_b = req_attrs_a;
 	}
 
-	req_attrs_a &= a->ce_mask;
-	req_attrs_b &= b->ce_mask;
+	if (ops->oo_id_attrs_get || ops->oo_id_attrs) {
+		req_attrs_a |= a->ce_mask;
+		req_attrs_b |= b->ce_mask;
+	} else {
+		req_attrs_a &= a->ce_mask;
+		req_attrs_b &= b->ce_mask;
+	}
 
 	/* Both objects must provide all required attributes to uniquely
 	 * identify an object */


lib/object.c
```diff
@@ -338,8 +338,13 @@ int nl_object_identical(struct nl_object *a, struct nl_object *b)
...
 	}
 
-	req_attrs_a &= a->ce_mask;
-	req_attrs_b &= b->ce_mask;
+	if (ops->oo_id_attrs_get || ops->oo_id_attrs) {
+		req_attrs_a |= a->ce_mask;
+		req_attrs_b |= b->ce_mask;
+	} else {
+		req_attrs_a &= a->ce_mask;
+		req_attrs_b &= b->ce_mask;
```
thom311, line 346: This doesn't seem right to me. I think the previous code is correct.

`nl_object_identical()` compares objects not fully, but whether their primary key attribute correspond.

whether `ops` is determining the key attributes via `ops->oo_id_attrs_get`, `ops->oo_id_attrs` or as `UINT64_MAX`, doesn't really matter -- except that `ops->oo_id_attrs_get` allows to determine the key attributes based on the object-instance (and not statically for all objects of the given type). Not setting `ops->oo_id_attrs` is merely a convenience shortcut to set it to `UINT64_MAX`. It does not imply any different behavior during the comparison.

I think the current code is correct. We determine the key attributes `req_attrs_a` (per object instance). Afterwards, we filter out the unset attrbitues (`&= a->ce_mask`) and compare only the set ones. The point of this `&=` is that if the objects don't have all required-attributes set, that they still can compare as identical as long as all their set attributes agree.

thom311:
in other words...

we have primary key attributes (`req_attrs_a` + `req_attrs_b`) which are used for comparing the ID of two objects.

Usually, we'd expect that both objects need all key attributes set. Then `&= a->ce_mask` has no effect, and we proceed to compare all the key attributes. 

In the case where the object has some key attributes unset, we still need to do something. Maybe such an incomplete object should never compare identical to another object. However, instead, even such an incomplete object can still compare identical to another object `b`, if-and-only-if that other object
- lacks the same key attributes
- has the same values for the present key attributes.

That is what `&= a->ce_mask` accomplishes.


There is a small twist, that with `req_attrs_a = ops->oo_id_attrs_get(a)`, the key attributes can depend on the object itself. So we also require that the set of key attributes is the same `if (req_attrs_a != req_attrs_b) return 0;`.

lib/object.c
voldymyr: You are right. I was partially confused by the name of the function _nl_object_identical()_ which made me think that it should make exhaustive object comparison, and partially by our code that was built around previous (erroneous) behavior of given function. All that directed me towards misconception so thanks a lot for taking the time explain functions conceived behavior. Feel free to close this PR.

PR #327, opened 2022-07-06
Use print() function in both Python 2 and Python 3

cclauss:
Python 2 died 917 days ago on 1/1/2020 so it is no longer available on several modern OS distributions.

Let's update the Python code so it has no Python 3 syntax errors but should still run on Python 2.

~Run [`psf/black`](https://github.com/psf/black#readme) on the three Python examples so they have standard formatting.~ -- Reverted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,17 @@ jobs:
             libtool-bin
 
       - name: Check out repository code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
+      - name: Setup Python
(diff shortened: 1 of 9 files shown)


thom311:
could you please squash the two commits? As long as the branch is WIP/on-review, it is preferable to rework it and rebase it. No need to first do something and revert it in a second commit. Thanks.

thom311:
> could you please squash the two commits? As long as the branch is WIP/on-review, it is preferable to rework it and rebase it. No need to first do something and revert it in a second commit. Thanks.

ah, you already did... never mind.

PR #328, opened 2022-08-12
route/vlan: fix error handling in 'lib/route/link/vlan.c'

kacperlud:


lib/route/link/vlan.c
```diff
@@ -313,7 +313,7 @@ static int vlan_put_attrs(struct nl_msg *msg, struct rtnl_link *link)
 		int i;
 
 		if (!(qos = nla_nest_start(msg, IFLA_VLAN_INGRESS_QOS)))
-			goto nla_put_failure;
+			return -NLE_MSGSIZE;
```
thom311, line 316: this fix is not right (but there is a bug that needs fixing).

The `NL_PUT*()` macros all `goto nla_put_failure`, so the function has multiple places where we jump to that label. And depending on from where we jump, we need to cancel 1 or 2 nestings.


I think there should be a `int nest_count;`, which gets incremented/decremented with `nla_nest_start()`/`nla_nest_end()`, to know how many times to cancel it.

kacperlud: Changed

thom311:
merged. Thank you!!

PR #332, opened 2022-10-13
route: add nexthop type

zstas:
The basic support for dumping nexthops and filling the cache.


lib/route/nh.c
```diff
@@ -0,0 +1,529 @@
...
+	return nhg;
+}
+
+static void rtnl_nh_grp_free(nl_nh_group_t *nhg)
+{
+	if (!nhg)
+		return;
+
+	if (nhg->ce_refcnt > 0)
+		NL_DBG(1, "Warning: Freeing NHG object in use...\n");
```
thom311, line 62: there is only one caller of the function, and that one already checks that the refcount is not positive. This code can never be reached.

Also, a logging about a bug is not great. Let's do `_nl_assert(nhg->ce_refcnt == 0);`

(note that the caller actually only checks that the ce_refcnt is `<= 0`. But it really can also never be negative, that would be a bug too, and the assert will be more strict than the caller).

Same for the check of `if (!nhg)`. This is a static function (called from a few lines below). You can easily see, that the condition can never be true. An `_nl_assert(nhg)` would be more appropriate.


`_nl_assert()` is usually disabled in release builds (unless the user builds with `CFLAGS=-DNL_MORE_ASSERTS=1000`, which is uncommon).

lib/route/nh.c
```diff
@@ -0,0 +1,529 @@
...
+		uint32_t	nh_id;
+	} __attribute__((packed)) nhkey;
+
+	nhkey_sz = sizeof(nhkey);
+	nhkey.nh_id = nexthop->nh_id;
+
+	*hashkey = nl_hash(&nhkey, nhkey_sz, 0) % table_sz;
+
+	NL_DBG(5, "nexthop %p key (nhid %d) keysz %d, hash 0x%x\n",
+				 nexthop, nhkey.nh_id, nhkey_sz, *hashkey);
```
thom311, line 137: I think a lot of the debug logging (all?) should be dropped. It doesn't seem useful when a library logs to stdout/stderr.

thom311 (changes requested):
thank you for your patch. I am sorry for the long time it took me to reply. If you lost interest, I understand :( Anyway. Some comments...

libnl-cli-3.sym
```diff
@@ -114,4 +114,6 @@ global:
 	nl_cli_alloc_cache_flags;
 	nl_cli_link_alloc_cache_flags;
 	nl_cli_link_alloc_cache_family_flags;
+	nl_cli_nh_alloc;
+	nl_cli_nh_alloc_cache;
```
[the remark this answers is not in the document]
zstas: done

libnl-route-3.sym
```diff
@@ -1267,4 +1267,18 @@ global:
...
+	rtnl_nh_get;
+	rtnl_nh_compare;
+	rtnl_nh_dump;
+	rtnl_nh_set_gateway;
+	rtnl_nh_get_gateway;
+	rtnl_nh_set_fdb;
+	rtnl_nh_set_fdb;
+	rtnl_nh_get_group_entry;
+	rtnl_nh_get_group_size;
+	rtnl_nh_get_id;
```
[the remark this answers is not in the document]
zstas: done

src/nl-nh-list.c
```diff
@@ -0,0 +1,58 @@
...
+"Usage: nl-nh-list [OPTIONS]... \n"
+"\n"
+"OPTIONS\n"
+" -h, --help                Show this help text.\n"
+" -v, --version             Show versioning information.\n"
+	);
+	exit(0);
+}
+
+int main(int argc, char *argv[])
```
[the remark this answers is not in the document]
zstas: done

include/linux-private/linux/nexthop.h
```diff
@@ -0,0 +1,104 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+#ifndef _LINUX_NEXTHOP_H
+#define _LINUX_NEXTHOP_H
+
+#include <linux/types.h>
```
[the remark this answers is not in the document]
zstas: done

lib/route/nh.c
zstas: done

lib/route/nh.c
```diff
@@ -0,0 +1,529 @@
...
+ * the operation succeeds, the resulting cache will contain a nexthop object for
+ * each nexthop configured in the kernel. If \c sk is NULL, returns 0 but the
+ * cache is still empty.
+ *
+ * @return 0 on success or a negative error code.
+ */
+static int rtnl_nh_alloc_cache_flags(struct nl_sock *sk, int family,
+				struct nl_cache **result, unsigned int flags)
+{
+	struct nl_cache * cache;
```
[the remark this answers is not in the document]
zstas: done

lib/route/nh.c
```diff
@@ -0,0 +1,529 @@
...
+		nexthop->ce_mask |= NH_ATTR_FLAG_GROUPS;
+	}
+
+	if (tb[NHA_FDB]) {
+		nexthop->ce_mask |= NH_ATTR_FLAG_FDB;
+	}
+
+	if (tb[NHA_GROUP]) {
+		list_len = 0;
+		uint8_t *data = nla_data (tb[NHA_GROUP]);
```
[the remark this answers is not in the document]
zstas: done

lib/route/nh.c
zstas: done

zstas:
H Thamas,

Thanks for the review. I tried to address all of them. Also I included another fix in the parser (in the previous patch only 2 group elements were allowed, I've increased it to 64). But unfortunately after rebasing github doesn't allow to compare the previous version with the current. But I've splitted the PR into 4 different commits (as you suggested), perhaps it makes the next review more easy.

I've been using the feature introduced in this patch for almost a year (for parsing fdb nexthops in linux), so I'd like to have this patch merged into upstream. So, I'm ready to improve this patch until you say it's good to go :)

include/netlink-private/types.h
```diff
@@ -254,6 +255,7 @@ struct rtnl_neigh
 	uint8_t		n_type;
 	struct nl_addr *n_lladdr;
 	struct nl_addr *n_dst;
+	uint32_t	n_nhid;
```
thom311, line 258: this needs to go the the previous commit (2 of 4). Otherwise, that commit does not compile.

lib/route/neigh.c
```diff
@@ -511,6 +520,9 @@ static void neigh_dump_line(struct nl_object *a, struct nl_dump_params *p)
 	if (n->ce_mask & NEIGH_ATTR_VLAN)
 		nl_dump(p, "vlan %d ", n->n_vlan);
 
+	if (n->ce_mask & NEIGH_ATTR_NHID)
+		nl_dump(p, "nhid %d ", n->n_nhid);
```
thom311, line 524: needs `"%u"` because `n->n_nhid` is `uint32_t`.

Technically, in plain C it might be that `uint32_t` is not the same as `unsigned int`, and thus `"%u"` would not be right. However, we rely on `sizeof(unsigned) >= sizeof(uint32_t)`, so `"%u"` is actually correct. To my knowledge, Linux kernel requires that `int` is at least 32bit(??).

include/netlink/route/neighbour.h
```diff
@@ -79,6 +79,9 @@ extern int			rtnl_neigh_get_vlan(struct rtnl_neigh *);
 extern void			rtnl_neigh_set_master(struct rtnl_neigh *, int);
 extern int			rtnl_neigh_get_master(struct rtnl_neigh *);
 
+extern void			rtnl_neigh_set_nhid(struct rtnl_neigh *, unsigned int);
+extern int			rtnl_neigh_get_nhid(struct rtnl_neigh *);
```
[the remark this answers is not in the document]
thom311: The setter should get a `uint32_t` argument (note that libnl3 headers already include `<stdint.h>`, so we are fine to use that in public headers.

Also, the getter needs to return a full `uint32_t`.
```c
uint32_t rtnl_neigh_get_nhid(struct rtnl_neigh *);
```
might work, but then there is no room for an error code.

I think, this needs to be
```c
int rtnl_neigh_get_nhid(struct rtnl_neigh *neight, uint32_t *out_val);
```
where it return 0 on success or a negative `NLE_*` errror code.

lib/route/neigh.c
```diff
@@ -1066,6 +1078,19 @@ int rtnl_neigh_get_master(struct rtnl_neigh *neigh) {
...
+{
+	neigh->n_nhid = nhid;
+	neigh->ce_mask |= NEIGH_ATTR_NHID;
+}
+
+int rtnl_neigh_get_nhid(struct rtnl_neigh *neigh) {
+	if (neigh->ce_mask & NEIGH_ATTR_NHID)
+		return neigh->n_nhid;
+	else
+		return -1;
```
thom311, line 1091: `return -NLE_NOATTR;`

thom311: can we do the error-paths first and return-early (not if/else)?
```c
    if (!(neigh->ce_mask & NEIGH_ATTR_NHID))
        return -NLE_NOATTR;
    return neigh->n_nhid;
```

include/netlink/route/neighbour.h
```diff
@@ -79,6 +79,9 @@ extern int			rtnl_neigh_get_vlan(struct rtnl_neigh *);
 extern void			rtnl_neigh_set_master(struct rtnl_neigh *, int);
 extern int			rtnl_neigh_get_master(struct rtnl_neigh *);
 
+extern void			rtnl_neigh_set_nhid(struct rtnl_neigh *, unsigned int);
```
thom311, line 82: does the patch lack a
```c
    if (tmpl->ce_mask & NEIGH_ATTR_NHID)
        NLA_PUT_U32(msg, NDA_NH_ID, tmpl->n_nhid);
```

?

otherwise, there is no point of having `rtnl_neigh_set_nhid()` in the public API.

Makefile.am
```diff
@@ -110,6 +110,7 @@ libnlinclude_netlink_route_HEADERS = \
 	include/netlink/route/neightbl.h \
 	include/netlink/route/netconf.h \
 	include/netlink/route/nexthop.h \
+	include/netlink/route/nh.h \
```
thom311, line 113: I added a script to automatically format new source files with `clang-format`. There is also a github action that tests the format, but it was broken. If you would rebase to latest `main`, you would see that the github test `clang-format` fails due to that.

If you would, please reformat the files. As the tool would tell you:
```
Fix it by running `"./tools/clang-format.sh"` using clang-format version 16.0.6 (Fedora 16.0.6-2.fc38)
Alternatively, run `./tools/clang-format-container.sh` to use a podman container named "libnl-code-format-f38".
```

---

If that is cumbersome for you (e.g. because you don't use Fedora 38 with the right clang-format version and don't have podman working), you can also ignore that, then I will reformat your patch before merging. Just ignore it.

libnl-cli-3.sym
```diff
@@ -121,3 +121,10 @@ global:
 	nl_cli_link_alloc_cache_flags;
 	nl_cli_link_alloc_cache_family_flags;
 } libnl_3;
+
+libnl_3_8 {
+global:
+	nl_cli_nh_alloc;
+	nl_cli_nh_alloc_cache;
+} libnl_3_2_28;
+
```
thom311, line 130: let's drop the empty line there.

libnl-route-3.sym
```diff
@@ -1286,4 +1286,18 @@ global:
...
```
thom311, line 1302: please sort the lines lexicographically.

include/netlink-private/types.h
```diff
@@ -323,6 +325,31 @@ struct rtnl_nexthop
...
+	uint8_t nh_scope;
+	uint8_t nh_protocol;
+	uint32_t nh_flags;
+
+	uint32_t nh_id;
+	uint32_t nh_group_type;
+	nl_nh_group_t * nh_group;
+	uint32_t nh_oif;
+	struct nl_addr * nh_gateway;
+	uint32_t nh_master;
```
thom311, line 350: some of those fields are unused (not yet implemented?). For example, `nh_protocol`.

It would seem clearer to just drop those. Otherwise, it prompts questions like whether `nh_clone()` needs to take care of those fields.

lib/route/nh.c
```diff
@@ -0,0 +1,526 @@
...
+	dst->nh_master = src->nh_master;
+	dst->ce_mask = src->ce_mask;
+
+	return 0;
+}
+
+static void nh_free(struct nl_object *obj)
+{
+	struct rtnl_nh *nh = nl_object_priv(obj);
+	nl_addr_put(nh->nh_gateway);
```
thom311, line 92: please add blank line before code.

lib/route/nh.c
```diff
@@ -0,0 +1,526 @@
...
+	if (nh->nh_group)
+		rtnl_nh_grp_put(nh->nh_group);
+}
+
+void rtnl_nh_put(struct rtnl_nh *nh)
+{
+	struct nl_object *obj = (struct nl_object *)nh;
+	nl_object_put(obj);
```
thom311, line 100: please add blank line before code.

lib/route/nh.c
```diff
@@ -0,0 +1,526 @@
...
+			    nl_nh_group_t **nvi)
+{
+	int cur = 0;
+	nl_nh_group_t *ret;
+
+	if (len <= 0)
+		return 0;
+
+	if (!(ret = rtnl_nh_grp_alloc(len)))
+		return -1;
```
thom311, line 189: `return -NLE_NOMEM;`

thom311 (changes requested):

include/netlink/route/neighbour.h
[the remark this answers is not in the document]
zstas: done

lib/route/neigh.c
zstas: done

lib/route/neigh.c
zstas: done

include/netlink/route/neighbour.h
zstas: I've added population the attribute.

lib/route/nh.c
zstas: done

lib/route/nh.c
zstas: done

lib/route/nh.c
zstas: done

include/netlink-private/types.h
zstas: done

libnl-route-3.sym
zstas: done

Makefile.am
zstas: after a couple of tries it worked with my clang-format-14

include/netlink-private/types.h
zstas: done

zstas:
I've addressed all the previous comments.
Also apologies for another last-minute change - since you mentioned nh_clone, I took another look and found that I wasn't copying nh_gateway and nh_group. Addressed that as well (in 3rd commit).

thom311 (changes requested):
Maybe you could take the patches from https://paste.centos.org/view/3c278476 (well, review them, and take the parts that you agree with).

It also requires you to rebase the branch again to latest `main` (because they use some new API from `main` branch).

zstas:
I intended to merge things from your patch today, but unfortunately the link is expired. Could you please share it one more time?

lib/route/neigh.c
```diff
@@ -721,6 +733,9 @@ static int build_neigh_msg(struct rtnl_neigh *tmpl, int cmd, int flags,
 	if (tmpl->ce_mask & NEIGH_ATTR_VLAN)
 		NLA_PUT_U16(msg, NDA_VLAN, tmpl->n_vlan);
 
+	if (tmpl->ce_mask & NEIGH_ATTR_NHID)
+		NLA_PUT_U32(msg, NDA_NH_ID, tmpl->n_nhid);
```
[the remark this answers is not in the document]
zstas: No, user-space must create a nexthop first (with a DST or a group, for vxlan also with flag `fdb`) and then create a neighbour entry with corresponding NH_ID. Dumping works the same way as it was created.

So, IMO it's better to skip that check and leave it to kernel.

thom311:
I didn't realize fpaste expires so fast. Sorry about that. Here again: https://paste.centos.org/view/bba2d453

I also just realized, that a change I did will cause a merge conflict when you rebase. Sorry about that. It should however be easy to resolved... Thanks.

zstas:
I rebased, applied all your changes and also removed nh_master attribute from `rtnl_nh` struct (as it wasn't used anywhere). Could you please take a look once again?

thom311:
merged as bc8573d035eeea37d53d715a342111176f29fb19.

Thank you for your efforts!!!

PR #333, opened 2022-10-17
route/link: prevent segfault in af_request_type()

TarEnethil:
`ac_request_type()` in `route/link.c` will segfault if the function pointer `ops->ao_override_rtm` has not been initialized for the given `af_type` parameter.

The bug was found while trying to call `rtnl_link_change()` for a link with family type AF_INET6. It seems this bug was introduced in 3c427d0.

Fix: check that `ao_override_rtm` is a valid pointer before calling it.


thom311:
merged to `main`. Thank you for your patch, and sorry for the long time...

PR #335, opened 2022-10-21
test-cache-mngr improvements

gobenji:
test-cache-mngr is very useful to reproduce and investigate problems in libnl.
This series adds options to tune the behavior of test-cache-mngr. Those
options were used to reproduce various bugs in the version of libnl used in
Cumulus Linux. Except for the last patch, the default behavior remains
unchanged.

### DIFF
--- a/tests/test-cache-mngr.c
+++ b/tests/test-cache-mngr.c
@@ -9,7 +9,7 @@
 
 static int quit = 0;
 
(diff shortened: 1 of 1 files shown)


thom311:
merged. Thank you for the patches!!

(and sorry for the very long response time).

PR #338, opened 2023-01-24
seg6_local: initial implementation, borrowed from linux kernel

kogdenko:


lib/route/seg6.c
```diff
@@ -0,0 +1,76 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+
+#include <stdbool.h>
+#include <linux/seg6.h>
+#include <linux/ipv6.h>
+#include <netlink/utils.h>
+#include <netlink-private/utils.h>
+
+bool seg6_validate_srh(struct ipv6_sr_hdr *srh, int len, bool reduced)
```
thom311, line 9: this function is taken from kernel sources, where it's licensed as GPL-2.0-or-later.

Here the license should be LGPL-2.1-only, which is incompatible.

The code seems not trivial enough to just ignore the problem. It also is not clear how to re-implement it, without looking at the original code (clean-room implementation). Maybe it would be possible, to ask the original authors whether they would agree to allow using their code under LGPL-2.1-only, so that it can be used in libnl3. The original authors possibly can be found with `git log -L'/seg6_validate_srh/,/^}$/:net/ipv6/seg6.c'` in the kernel sources.

Makefile.am
```diff
@@ -243,6 +243,9 @@ noinst_HEADERS = \
 	include/linux-private/linux/lwtunnel.h \
 	include/linux-private/linux/mpls.h \
 	include/linux-private/linux/mpls_iptunnel.h \
+	include/linux-private/linux/seg6_local.h \
+	include/linux-private/linux/seg6.h \
+	include/linux-private/linux/seg6_iptunnel.h \
```
thom311, line 248: please sort lines alphabetically.

lib/route/seg6.c
```diff
@@ -0,0 +1,76 @@
...
+
+void seg6_dump_srh(struct nl_dump_params *dp, struct ipv6_sr_hdr *srh)
+{
+	int i, offset;
+	char addr[INET6_ADDRSTRLEN];
+	struct sr6_tlv_hmac *tlv;
+
+	nl_dump(dp, "segs %d [ ", srh->first_segment + 1);
+	for (i = srh->first_segment; i >= 0; i--) {
+		nl_dump(dp, "%s", _nl_inet_ntop(AF_INET6, &srh->segments[i], addr));
```
thom311, line 67: why does it print in reverse order?

lib/route/seg6.c
```diff
@@ -0,0 +1,76 @@
...
+	}
+	nl_dump(dp, "] ");
+	if (sr_has_hmac(srh)) {
+		offset = ((srh->hdrlen + 1) << 3) - 40;
```
thom311, line 71: where does the 40 come from? The entire expression is rather hard to understand. It should be simplified, possibly using a define for non-obvious things. Or at least a code comment, that explains where thsoe numbers come from.

lib/route/seg6.c
thom311: it also needs a code comment, that indicates where the code was copied from.

lib/route/nh_encap_seg6_local.c
```diff
@@ -0,0 +1,1161 @@
...
+	unsigned long attrs;
+	struct seg6_local_lwt *slwt;
+
+	slwt = get_seg6_local_slwt(nh);
+	if (slwt == NULL)
+		return 0;
+
+	attrs = slwt->desc->attrs | slwt->parsed_optattrs;
+
+	return attrs & SEG6_F_ATTR(attr);
```
thom311, line 1138: `SEG6_F_ATTR(attr)` does a bitshift based on `attr`. attr is user provided input, let's check that the `attr` argument is suitably small (and non-negative), to not cause undefined behavior.

```C
if (attr < 0 or attr >= sizeof(attr)*8) return -NLE_INVAL;
```

But more importantly, please return a negative error code
- `attr` is invalid -> `-NLE_INVAL`
- `if (!slwt)` -> `-NLE_NOATTR`

otherwise, return either 0 or 1, via `return !!(attrs & SEG6_F_ATTR(attr));`

A doctext explaining the purpuse/use of the function and behavior (return values) would also be necessary.

lib/route/nh_encap_seg6_local.c
```diff
@@ -0,0 +1,1161 @@
...
+}
+
+static int put_nla_srh(struct nl_msg *msg, struct seg6_local_lwt *slwt)
+{
+	struct ipv6_sr_hdr *srh;
+	struct nlattr *nla;
+	int len;
+
+	srh = slwt->srh;
+	len = (srh->hdrlen + 1) << 3;
```
thom311, line 254: this expression is used several times, and it's not clear why it's calculated that way. Please add a C function for it (and possibly a code comment explaining the calculation).

lib/route/nh_encap_seg6_local.c
```diff
@@ -0,0 +1,1161 @@
...
+	memcpy(nla_data(nla), srh, len);
+
+	return 0;
+}
+
+static int cmp_nla_srh(struct seg6_local_lwt *a, struct seg6_local_lwt *b)
+{
+	int len = (a->srh->hdrlen + 1) << 3;
+
+	if (len != ((b->srh->hdrlen + 1) << 3))
```
thom311, line 269: `if (a->srh->hdrlen != b->srh->hdrlen)`

thom311 (changes requested):
I am sorry for my late reply. Thank you for you efforts.

configure.ac
```diff
@@ -29,9 +29,9 @@ m4_define([libnl_git_sha], [m4_esyscmd([ ( [ -d ./.git/ ] && [ "$(readlink -f ./
 # 3. Programs may need to be changed, recompiled, relinked in order to use
 #    the new version. Bump current, set revision and age to 0.
 
-m4_define([libnl_lt_current],    [226])
+m4_define([libnl_lt_current],    [227])
 m4_define([libnl_lt_revision],   [0])
-m4_define([libnl_lt_age],        [26])
+m4_define([libnl_lt_age],        [27])
```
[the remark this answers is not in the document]
kogdenko: done

lib/route/seg6.c
kogdenko: Code comments added.

Not only seg6_validate_srh() is trivial, but some other funcitions:
**Author: David Lebrun <david.lebrun@uclouvain.be>**
lib/route/nh_encap_seg6.c: seg6_encap_parse_msg,
lib/route/seg6.c: seg6_validate_srh, seg6_dump_srh
lib/route/nh_encap_seg6_local.c: parse_nla_srh, put_nla_srh, parse_nla_action, seg6_local_encap_parse_msg
**Author: Andrea Mayer <andrea.mayer@uniroma2.it>**
lib/route/nh_encap_seg6_local.c: parse_nla_counters, seg6_local_fill_nla_counters, put_nla_counters, seg6_chk_next_csid_cfg, seg6_parse_nla_next_csid_cfg, parse_nla_flavors, put_nla_flavors, parse_nla_optional_attrs
**Author: Mathieu Xhonneux <m.xhonneux@gmail.com>**
lib/route/nh_encap_seg6_local.c: parse_nla_bpf, put_nla_bpf

I think after resolving other issues with this PR we should ask those people to sign commit

libnl-route-3.sym
```diff
@@ -1284,3 +1284,13 @@ global:
...
+
+libnl_3_9 {
+	rtnl_route_nh_get_encap_type;
+	rtnl_route_nh_get_encap_seg6_mode;
+	rtnl_route_nh_get_encap_seg6_srh;
+	rtnl_route_nh_get_encap_seg6_local_action;
+	rtnl_route_nh_has_encap_seg6_local_attr;
+	rtnl_route_nh_get_encap_seg6_local_table;
+	rtnl_route_nh_get_encap_seg6_local_vrftable;
+} libnl_3_8;
```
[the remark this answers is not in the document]
kogdenko: done

lib/route/seg6.c
kogdenko: Code copied from iproute2 (Comment was added). I think dump should be similar with ip utility

lib/route/nh_encap_seg6_local.c
kogdenko: Add common macros: IPV6_EXTHDR_LEN

lib/route/seg6.c
kogdenko: 40 changed to sizeof(..)

include/netlink/route/nexthop.h
```diff
@@ -61,11 +61,23 @@ extern int		rtnl_route_nh_str2flags(const char *);
...
+
 extern int		rtnl_route_nh_encap_mpls(struct rtnl_nexthop *nh,
 						 struct nl_addr *addr,
 						 uint8_t ttl);
 extern struct nl_addr *	rtnl_route_nh_get_encap_mpls_dst(struct rtnl_nexthop *);
 extern uint8_t		rtnl_route_nh_get_encap_mpls_ttl(struct rtnl_nexthop *);
+
+
+extern int		rtnl_route_nh_get_encap_seg6_mode(struct rtnl_nexthop * );
+extern int		rtnl_route_nh_get_encap_seg6_srh(struct rtnl_nexthop *, void **);
```
thom311, line 74: the `void**` here is really the `struct ipv6_sr_hdr`. OK, I think we can do this, but then it should be
```
struct ipv6_sr_hdr;
extern int		rtnl_route_nh_get_encap_seg6_srh(struct rtnl_nexthop *, const struct ipv6_sr_hdr **);
```

also, the user wouldn't know the actual size (`IPV6_EXTHDR_LEN`), so the size should be returned by `rtnl_route_nh_get_encap_seg6_srh()`.


The alternative (maybe better) would be to add a `struct rtnl_ipv6_sr_hdr` that is fully under libnl control, and which possibly exposes a more convenient layout of fields (e.g. the number of segments as a fields of the struct).

include/netlink/route/nexthop.h
```diff
@@ -61,11 +61,23 @@ extern int		rtnl_route_nh_str2flags(const char *);
...
+
+extern int		rtnl_route_nh_get_encap_seg6_local_action(struct rtnl_nexthop *);
+extern int		rtnl_route_nh_has_encap_seg6_local_attr(struct rtnl_nexthop *, int);
```
thom311, line 77: The argument "attr" is an enum value from kernel headers, like `SEG6_LOCAL_FLAVORS`. If the user wants to use this API, they would have to include the kernel header. And since `SEG6_LOCAL_FLAVORS` was only added a short time ago, they might need to use the numerical value. OK, that's doable, but ugly.

How is this function to be used?

lib/route/nh_encap_seg6_local.c
```diff
@@ -0,0 +1,1202 @@
...
+	int i, err;
+
+	NLA_PUT_U32(msg, SEG6_LOCAL_ACTION, slwt->action);
+
+	attrs = slwt->desc->attrs | slwt->parsed_optattrs;
+
+	for (i = SEG6_LOCAL_SRH; i < SEG6_LOCAL_MAX + 1; i++) {
+		if (attrs & SEG6_F_ATTR(i)) {
+			param = &seg6_action_params[i];
+			err = param->put(msg, slwt);
```
thom311, line 1046: put constructs a netlink message to send to kernel, right?

E.g. there is `rtnl_route_nh_get_encap_seg6_local_table()`, but there is no setter. So you cannot create a libnl object with the local-table set, and thus you cannot set the SEG6_LOCAL_TABLE attribute (and the put code is unusable).

Or how this usable?

lib/route/nh_encap_seg6.c
```diff
@@ -0,0 +1,166 @@
...
+	case SEG6_IPTUN_MODE_ENCAP_RED:
+		break;
+	case SEG6_IPTUN_MODE_L2ENCAP_RED:
+		break;
+	default:
+		return -NLE_INVAL;
+	}
+
+	/* verify that SRH is consistent */
+	if (!seg6_validate_srh(tuninfo->srh, tuninfo_len - sizeof(*tuninfo), false))
```
thom311, line 69: why is this a good idea?

Kernel clearly needs to validate, the netlink messages that it gets from user-space. But here we get netlink messages from kernel. We trust kernel to a certain extend to do the right thing, and if it doesn't, shouldn't we try to make the best of it, and interpret it the best we can?

The check like `if (IPV6_EXTHDR_LEN(srh->hdrlen) != len)` is of course important, to ensure that we have the expected buffer size. But why would libnl care about `if (srh->type != IPV6_SRCRT_TYPE_4)` ?

I don't understand `seg6_validate_srh()`? Why is this so complicated? I wouldn't mind, but `rtnl_route_nh_get_encap_seg6_srh()` exposes this complexity to the user (by giving access to that struct, which seems complex to interpret). Is the user expected to re-implment how to interpret a `struct ipv6_sr_hdr`?

thom311 (approved):
I pushed some fixup commits to `th/pr/338` branch. Maybe you can review and consider them?

include/netlink/route/nexthop.h
kogdenko: done

include/netlink/route/nexthop.h
kogdenko: In my case:
```
if (rtnl_route_nh_has_encap_seg6_local_attr(nh, SEG6_LOCAL_TABLE)) {
    table_id = rtnl_route_nh_get_encap_seg6_local_table(nh);
} else if (rtnl_route_nh_has_encap_seg6_local_attr(nh, SEG6_LOCAL_VRFTABLE)) {
    table_id = rtnl_route_nh_get_encap_seg6_local_vrftable(nh);
}
```

lib/route/nh_encap_seg6_local.c
kogdenko: I do not implement setter for SEG6_LOCAL_TABLE - thought it is OK, because not all attributes has setters (rtnl_route_nh_get_encap_mpls_ttl for example)
Do you think we need implement it now?

thom311:
I merged the first commit to `main` as aa7353fde8532355e2582dc99d06855b94e96cf9.

I rebased (and squashed) the other commits. They are now at `th/pr/338` branch. Would this be OK for you?

I think, we would need to reach out to the people who wrote the original code, whether they agree to the LGPL-2.1-only licensing (https://github.com/thom311/libnl/pull/338#discussion_r1288917347). Can that be done?


Thank you.

thom311:
this pull request is stale.

I think the latest version can still be found at `th/pr/338` branch (commit 8123fca8b6dd113e22d4d2e0bed075db29db22d5).

I will close this for now. If anybody want to push this forward, please pick up the work. Thank you!!

PR #341, opened 2023-03-19
macsec: Fix set offload capability for MACsec

emeel97:
Update macsec headers in sync with the latest Linux headers and enhance the macsec offload function in the libnl library to enable offloading all types of macsec offload.

### DIFF
--- a/lib/route/link/macsec.c
+++ b/lib/route/link/macsec.c
@@ -652,9 +652,6 @@ int rtnl_link_macsec_set_offload(struct rtnl_link *link, uint8_t offload)
 
 	IS_MACSEC_LINK_ASSERT(link);
 
-	if (offload > 1)
-		return -NLE_INVAL;
-
 	info->offload = offload;
 	info->ce_mask |= MACSEC_ATTR_OFFLOAD;
 


lib/route/link/macsec.c
```diff
@@ -652,7 +652,7 @@ int rtnl_link_macsec_set_offload(struct rtnl_link *link, uint8_t offload)
 
 	IS_MACSEC_LINK_ASSERT(link);
 
-	if (offload > 1)
+	if (offload > MACSEC_OFFLOAD_MAX)
```
thom311, line 655: hm. Maybe such validation is not useful to begin with.

I mean, the library trying to be helpful to the user when they call set-offload, but is it really helpful? Eventually the user will send the netlink message to kernel, and whether the message is valid depends on kernel.

- you might run against an older kernel, and then -- even if libnl3 was compiled against newer kernel headers, and no failure was emitted -- the request will fail.
- or you might run against a newer kernel that would support another offload value, but libnl3 would require a patch to make it working.

Also, the user in general must have somewhat of an understanding which values are suitable to pass to `rtnl_link_macsec_set_offload()`. If they pass valid values, the validation is useless at best (worst, it rejects a valid value). If they pass invalid values, then nothing is going to make that work, and kernel will not behave as desired.

thom311 (changes requested):
sorry for the late reply. What do you think about dropping this check altogether? Would you be willing to send a patch? Thanks.

lib/route/link/macsec.c
emeel97: I totally agree I basically used this library when I tried adding support for macsec offload in wpa_supplicant and it took me a while to figure out that i failed because of this check, I was passing a valid value (kernel wise) but it was getting blocked but the library. I can send a patch to drop this check.
Do I drop the kernel headers sync patch?

thom311: yes, please update the pull request to drop the check altogether (you can just `git push --force` and re-use this pull request).

About the kernel header sync. We can do it, but also can not. We usually do it when we need something new. As you prefer, both is fine.

Thank you!

emeel97: ack & done

thom311:
merged to `main`. Thank you for the patch, and your persistence!!

PR #342, opened 2023-03-29
route/link: add accessor API for IPv6 DEVCONF

abuibrahim:
Add accessors similar to rtnl_link_inet_{get,set}_conf().


thom311:
Thanks for the patch.

I merged it, but then I did a follow-up. The followup removes the "set-conf" function again. It seems not useful to me.

Please review those patches, and if you have anything to add/comment/correct, please do.

Thank you.

PR #344, opened 2023-04-27
lib/handlers: add attribute acccess for struct nl_cb cb.cb_refcnt

precla:
problem:
`nl_cb_put()` decreases the cb.cb_refcnt and if it reaches 0, it will free(cb). `free()` does not set `cb = NULL`, which means that if `cb.cb_refcnt > 1`, we have to call `nl_cb_put()` again until `cb.cb_refcnt = 0`

doing it with, for example, `while(cb) { nl_cb_put(cb); }` won't work becasue `free()` is not doing `cb == NULL` so we never know when `cb_refcnt == 0`.

access to `cb->cb_refcnt` is not possible from outside libnl, one has to
include `libnl/include/netlink-private/types.h` which is not trivial.

solution:
add this function to be able to cleanup properly if `cb.cb_refcnt > 1` so one can do, for example:
`while(cb && nl_cb_get_refcnt(cb) > 0) { nl_cb_put(cb); }`

### DIFF
--- a/include/netlink/handlers.h
+++ b/include/netlink/handlers.h
@@ -117,6 +117,8 @@ extern struct nl_cb *	nl_cb_clone(struct nl_cb *);
 extern struct nl_cb *	nl_cb_get(struct nl_cb *);
 extern void		nl_cb_put(struct nl_cb *);
 
+extern int		nl_cb_get_refcnt(const struct nl_cb *cb);
+
 extern int  nl_cb_set(struct nl_cb *, enum nl_cb_type, enum nl_cb_kind,
 		      nl_recvmsg_msg_cb_t, void *);
 extern int  nl_cb_set_all(struct nl_cb *, enum nl_cb_kind,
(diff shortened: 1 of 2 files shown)


thom311:
sorry for the late reply.

---

But this doesn't look like the right solution.

Like most objects in libnl, `struct nl_cb` uses reference counting (`nl_cb_get()` and `nl_cb_put()`). You must *always* know whether you own a reference to the structure, and release that reference at some point.

For example, if you call `nl_cb_clone()` or `nl_cb_get()` you get ownership of one reference to the struct. You *MUST* eventually call `nl_cb_put()` to return that reference, but only that single reference that you own. You don't know whether anybody else owns a reference (nor should you need to know).

In the frequent case, where all the references to an object are owned by your code (and you are aware how many references there are), when you return all those references, that object will be freed. But possibly some other part of the code owns references too, and will later call `nl_cb_put()`. You must not free those references, otherwise, that other part of the code will have a use-after free error.

In general, reference counted API does not provide a way to look at the reference count (except for debugging). The use case described in the commit message, seems like something that should not be done.

I am closing this for that reason. Sorry.

precla:
Thanks for the clarification!

PR #345, opened 2023-05-18
docs: rtnl_link_put() 'releases' instead of 'returns'

precla:


thom311:
when looking at 
```bash
git grep -B10 'void \+\(rtnl\|nl\|xfrmnl\)_[a-zA-Z0-9_]\+put *(.*[^;]$'
```
output, we see that the doctext description for put-functions is not consistent.

I really don't mind whether this uses "release" or "return" when talking about the unref/put functions, but when a patch comes along to address that, it would be nice to attempt to find a common wording.

I will merge this patch, because for `nl_object_put()` the comment says "Release a reference from an object". So changing `rtnl_link_put()` to say "Release a link object reference" is not obviously wrong. But it is still somewhat consistent wording...

Merged as https://github.com/thom311/libnl/commit/4c4e614b783a9bedac06e60459ddb41f3284525d

Thanks for the patch. Efforts to unify the inconsistencies are welcome :)

Issue #343, opened 2023-04-13
TC action stats are not updated


dcaratti:
A program calling `rtnl_cls_alloc_cache()` updates the TC action list, but stats are not updated. They are present in the netlink message created by kernel, but as `cls_msg_parser()`  calls `rtnl_act_parse()` the netlink attribute of TC statistics is never parsed.

Here is an example from bmon:
```
(gdb) bt
#0  gact_msg_parser (tc=0x48dcb0, data=0x48dde0) at lib/route/act/gact.c:26
#1  0x00007ffff7e92544 in rtnl_act_parse (head=head@entry=0x48db70, tb=<optimized out>) at lib/route/act.c:447
#2  0x00007ffff7e92763 in flower_msg_parser (tc=<optimized out>, data=0x48db70) at lib/route/cls/flower.c:75
#3  0x00007ffff7ec66db in rtnl_tc_msg_parse (n=<optimized out>, tc=0x48da40) at lib/route/tc.c:175
#4  0x00007ffff7e936e6 in cls_msg_parser (ops=<optimized out>, who=<optimized out>, nlh=0x48d930, pp=0x7fffffffd530) at lib/route/cls.c:452
#5  0x00007ffff7f061a2 in nl_cache_parse (params=<optimized out>, nlh=<optimized out>, who=<optimized out>, ops=<optimized out>) at lib/cache.c:989
#6  nl_cache_parse (ops=0x7ffff7ef14e0 <rtnl_cls_ops.lto_priv>, who=0x48c718, nlh=0x48d930, params=0x7fffffffd530) at lib/cache.c:979
#7  0x00007ffff7f061ef in update_msg_parser (msg=<optimized out>, arg=<optimized out>) at lib/cache.c:665
#8  0x00007ffff7f0ae21 in nl_cb_call (msg=<optimized out>, type=<optimized out>, cb=<optimized out>) at ./include/netlink-private/netlink.h:139
#9  recvmsgs (cb=0x48d680, sk=0x423d90) at lib/nl.c:1000
#10 nl_recvmsgs_report (sk=0x423d90, cb=0x48d680) at lib/nl.c:1051
#11 0x00007ffff7f0b5dd in nl_recvmsgs (sk=sk@entry=0x423d90, cb=cb@entry=0x48d680) at lib/nl.c:1075
[...]

PR #346, opened 2023-05-23
add support for TC action statistics

dcaratti:
https://github.com/thom311/libnl/issues/343


lib/route/act.c
```diff
@@ -435,6 +442,43 @@ int rtnl_act_parse(struct rtnl_act **head, struct nlattr *tb)
...
+			err = nla_parse_nested(tb3, TCA_STATS_MAX, tb2[TCA_ACT_STATS],
+					       tc_act_stats_policy);
+			if (err < 0)
+				return err;
+
+			if (tb3[TCA_STATS_BASIC]) {
+				struct gnet_stats_basic *bs;
+
+				bs = nla_data(tb3[TCA_STATS_BASIC]);
+				tc->tc_stats[RTNL_TC_BYTES] = bs->bytes;
```
thom311, line 457: I think the netlink attribute is only guaranteed to have an alignment of 4 (`sizeof(uint32_t)`). Maybe (just maybe), we also assume that `sizeof(int) == 4`, so it would also be valid for `int`/`unsigned` fields (but I think we should not assume that).

However, `bs->bytes` and `struct gnet_stats_rate_est64` are uint64. We need to memcpy them.

Maybe a simpler solution is to
```C
    struct gnet_stats_basic bs;

    memcpy(&bs, nla_data(tb3[TCA_STATS_BASIC]), sizeof(bs));
    tc->tc_stats[RTNL_TC_BYTES] = bs.bytes;
```
then we don't need to reason about each field individually, which type they are and when/how to memcpy.

thom311 (changes requested):
thank you for the patch.

could you please also rebase to latest `main` branch? If fixed the failure in the github actions, and it would be interesting to see it pass.

dcaratti:
> thank you for the patch.
> 
> could you please also rebase to latest `main` branch? If fixed the failure in the github actions, and it would be interesting to see it pass.

sure, I will! thanks for looking at this!

lib/route/act.c
dcaratti: for this specific attribute, nla_data is 4 bytes alignment and payload is padded to multiple of 8 bytes. If I well get the problem, it's about unaligned access when doing cast + assignment inside the netlink payload. I see this is done elsewhere in libnl3 - for [qdisc stats](https://github.com/thom311/libnl/blob/main/lib/route/tc.c#L110) - but yes,  I can change it (though I'm unsure I understood what kind of problem it can solve).

Thanks for reviewing!

thom311: the problem to avoid, is unaligned memory access (in this case, reading a uint64 integer).

As netlink attributes have a guaranteed alignment of "only" 4 bytes, I think it can happen that  `nla_data(tb3[TCA_STATS_BASIC])` starts at an address with `addresss%8==4`. Or why not?

I don't understand why that would *not* be a problem. I think https://github.com/thom311/libnl/blob/main/lib/route/tc.c#L110 is a bug. Or why not?

dcaratti: ... understood. Done, just force-pushed a v2 where I rebased and converted parser of `TCA_STATS_BASIC` and `TCA_STATS_RATE_EST64` to copy on a local struct allocated on the stack. The other 2 stats, namely `TCA_STATS_QUEUE` and `TCA_STATS_RATE_EST` have only 32-bit wide members, so all accessed should already be aligned to 4-bytes.

thom311:
merged. Thank you!!

PR #349, opened 2023-06-30
route/bond: Add support for link_info for bond

mallikarjunnemagoudar:
following API has been added to bonding
    bond_info_alloc
    bond_info_free
    rtnl_link_bond_set_mode
    rtnl_link_bond_set_activeslave
    bond_put_attrs

### DIFF
--- a/include/netlink/route/link/bonding.h
+++ b/include/netlink/route/link/bonding.h
@@ -25,6 +25,9 @@ extern int	rtnl_link_bond_enslave(struct nl_sock *, struct rtnl_link *,
 extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
 extern int	rtnl_link_bond_release(struct nl_sock *, struct rtnl_link *);
 
+extern void	rtnl_link_bond_set_mode(struct rtnl_link *link, uint8_t mode);
+extern void	rtnl_link_bond_set_activeslave(struct rtnl_link *link, int active_slave);
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/route/link/bonding.c
+++ b/lib/route/link/bonding.c
@@ -19,6 +19,109 @@
 #include <netlink/route/link/bonding.h>
 #include <netlink-private/route/link/api.h>
 
+#define BOND_HAS_MODE		(1 << 0)
+#define BOND_HAS_ACTIVE_SLAVE	(1 << 1)
+
+struct bond_info {
+	uint8_t bn_mode;
+	uint32_t ifindex;
+	uint32_t bn_mask;
+};
+
+static int bond_info_alloc(struct rtnl_link *link)
+{
+	struct bond_info *bn;
+
+	if (link->l_info)
+		memset(link->l_info, 0, sizeof(*bn));
+	else {
+		bn = calloc(1, sizeof(*bn));
+		if (!bn)
+			return -NLE_NOMEM;
+
+		link->l_info = bn;
+	}
+
+	return 0;
+}
+
+static void bond_info_free(struct rtnl_link *link)
+{
+	_nl_clear_free(&link->l_info);
+}
+
+static int bond_put_attrs(struct nl_msg *msg, struct rtnl_link *link)
+{
+	struct bond_info *bn = link->l_info;
+	struct nlattr *data;
+
+	data = nla_nest_start(msg, IFLA_INFO_DATA);
+	if (!data)
+		return -NLE_MSGSIZE;
+	if (bn->bn_mask & BOND_HAS_MODE)
+		NLA_PUT_U8(msg, IFLA_BOND_MODE, bn->bn_mode);
+
+	if (bn->bn_mask & BOND_HAS_ACTIVE_SLAVE)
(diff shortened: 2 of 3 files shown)


include/netlink/route/link/bonding.h
```diff
@@ -12,6 +12,9 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+extern void rtnl_link_set_bondmode(struct rtnl_link *, uint8_t);
+
+extern void rtnl_link_set_bondactiveslave(struct rtnl_link *, int);
```
thom311, line 17: I think the name is not right. Compare to `rtnl_link_bridge_set_vlan_filtering()` which is bridge-specifc, and the name makes that clear. `rtnl_link_set_bondmode()` looks like `rtnl_link_set_carrier()`, which operates on the general link object.

I think it should be named `rtnl_link_bond_set_*()`.

Also, please move the new setters to the bottom, not before `rtnl_link_bond_alloc()`.

lib/route/link/bonding.c
```diff
@@ -19,6 +19,69 @@
...
+
+	return 0;
+}
+
+static void bond_free(struct rtnl_link *link)
+{
+	struct bond_info *bn = link->l_info;
+
+	free(bn);
+	link->l_info = NULL;
```
thom311, line 56: `_nl_clear_free(&link->l_info);`

lib/route/link/bonding.c
```diff
@@ -19,6 +19,69 @@
...
+/**
+ * Set active slave
+ * @arg link            Link object
+ * @arg active          ifindex of active slave
+ *
+ */
+void rtnl_link_set_bondactiveslave(struct rtnl_link *link, int active_slave)
+{
+        struct bond_info *bn = link->l_info;
+        bn->ifindex = active_slave;
```
thom311, line 68: I think we should check that the link is of the expected type. See for example `rtnl_link_bridge_set_vlan_filtering()` which asserts that the type is as expected.

lib/route/link/bonding.c
```diff
@@ -19,6 +19,69 @@
...
+        bn->bn_mask |= BOND_HAS_ACTIVE_SLAVE;
```
thom311, line 69: the code should be indented with tabs.

lib/route/link/bonding.c
```diff
@@ -19,6 +19,69 @@
...
+/**
+ * Set bond mode
+ * @arg link            Link object
+ * @arg mode            bond mode
+ *
+ */
+void rtnl_link_set_bondmode(struct rtnl_link *link, uint8_t mode)
+{
+	struct bond_info *bn = link->l_info;
+	bn->bn_mode = mode;
```
thom311, line 81: please a blank line between variable definitions and the code.

include/netlink/route/link/bonding.h
```diff
@@ -12,6 +12,9 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+extern void rtnl_link_set_bondmode(struct rtnl_link *, uint8_t);
```
thom311, line 15: API needs to be added to `libnl-route-3.sym` (at the `libnl_3_8` section at bottom of the file)

thom311 (changes requested):

lib/route/link/bonding.c
```diff
@@ -19,6 +19,108 @@
...
+		NLA_PUT_U8(msg, IFLA_BOND_MODE, bn->bn_mode);
+
+	if (bn->bn_mask & BOND_HAS_ACTIVE_SLAVE)
+		NLA_PUT_U32(msg, IFLA_BOND_ACTIVE_SLAVE, bn->ifindex);
+
+	nla_nest_end(msg, data);
+
+nla_put_failure:
+
+	return 0;
```
thom311, line 71: This can only be reached if the `NLA_PUT*()` macros fail, because there is not enough space in the netlink message buffer. In fact, it leaves the message broken, because we don't even call `nla_nest_cancel()`.

This means, we really need to return an error code here. When looking at `git grep -A5 nla_put_failure:`, I think this indicates that it should be
```c
nla_put_failure:
    nla_nest_cancel(msg, data);
    return -NLE_MSGSIZE;
```

thom311 (changes requested):
thanks for v2. Almost there :)

lib/route/link/bonding.c
mallikarjunnemagoudar: Thanks for the detailed explanation. Taken care now.

libnl-route-3.sym
```diff
@@ -1284,4 +1284,6 @@ global:
 	rtnl_link_bridge_set_vlan_protocol;
 	rtnl_link_bridge_set_vlan_stats_enabled;
 	rtnl_link_inet6_get_conf;
+	rtnl_link_bond_set_mode;
+	rtnl_link_bond_set_activeslave;
```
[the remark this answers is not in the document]
mallikarjunnemagoudar: Taken care.

thom311 (approved):

thom311:
merged to `main`. Thank you for the patch!!!

PR #359, opened 2023-08-30
python: Use correct decorator syntax in HTBQdisc

dpward:
Fixes: 87d370912ca8 ("netlink.nlattr re-implemented in more pythonic way")


thom311:
merged to `main`. Thank you!!

PR #360, opened 2023-08-31
Makefile.am: avoid use of non-portable echo arguments

thesamesam:
This fixes tests with a non-bash shell as /bin/sh (in this case, dash) which does not support `echo -e`. echo itself is portable, but not echo with any arguments.

Use `printf` instead.


thom311:
merged to main. Thank you!

thesamesam:
Thanks!

PR #366, opened 2023-11-27
Addtional leak removal and couple of improvements for xfrm code

spellingmistake:
Hi Thomas, these patches started out on an earlier version of libnl. When rebasing them to main (is renaming master to main RedHat policy?) I noticed that one change (fixing the leak in sa.c) was already applied (differently). I decided to stick with my changes. The introduced macros also add the `nl_addr_put` but also to sp.c.
The other two changes aren't particularly important., but fix some potential issues when using libnl.


lib/xfrm/sp.c
```diff
@@ -1338,11 +1338,25 @@ int xfrmnl_sp_set_userpolicy_type (struct xfrmnl_sp* sp, unsigned int type)
...
 {
-	nl_list_add_tail(&utmpl->utmpl_list, &sp->usertmpl_list);
-	sp->nr_user_tmpl++;
-	sp->ce_mask |= XFRM_SP_ATTR_TMPL;
+	if (MAX_TMPLS > sp->nr_user_tmpl) {
+		nl_list_add_tail(&utmpl->utmpl_list, &sp->usertmpl_list);
+		sp->nr_user_tmpl++;
+		sp->ce_mask |= XFRM_SP_ATTR_TMPL;
+		return 0;
+	}
```
thom311, line 1332: An existing function cannot change API in this way (e.g. change a `void` return to `int`). 

Also, is this useful? OK, the user surely doesn't want to configure more than 6 templates (as kernel won't reject it and it doesn't work). But currently the user can set that up in libnl and get a failure later. Now, they will get a failure during `xfrmnl_sp_add_usertemplate`. Does that solve any problem? It won't work for the user either way, and they are well advised to count to 6 themselves.

What if kernel tomorrow supports more than 6 templates?

Just let user-space try to configure the templates they want. If kernel thinks there are too many, it rejects it. That is fine behavior.

thom311, line 1332: An existing function cannot change API in this way (e.g. change a `void` return to `int`). 

Also, is this useful? OK, the user surely doesn't want to configure more than 6 templates (as kernel won't reject it and it doesn't work). But currently the user can set that up in libnl and get a failure later. Now, they will get a failure during `xfrmnl_sp_add_usertemplate`. Does that solve any problem? It won't work for the user either way, and they are well advised to count to 6 themselves.

What if kernel tomorrow supports more than 6 templates?

Just let user-space try to configure the templates they want. If kernel thinks there are too many, it rejects it. That is fine behavior.

include/netlink/netlink.h
```diff
@@ -32,6 +32,49 @@
...
+	} else { \
+		_addr = nl_addr_build((src)->family, \
+				&(src)->what ## addr.a6, \
+				sizeof((src)->what ## addr.a6)); \
+	} \
+	nl_addr_set_prefixlen(_addr, (src)->prefixlen_ ## what); \
+	xfrmnl_sel_set_ ## what ## addr((dst), _addr); \
+	/* xfrmnl_sel_set_*addr increases the refcount; without this addr would leak */ \
+	nl_addr_put(_addr); \
+	xfrmnl_sel_set_prefixlen_ ## what((dst), (src)->prefixlen_ ## what); \
```
thom311, line 52: I am sorry, such macros don't seem preferable to me.

Optimally, macros try to be very-function-like, that is:
- wrap in `do { } while (0)`
- evaluate all macro arguments exactly once (assigning to a local variable).
- avoid building names like `xfrmnl_sel_set_prefixlen_ ## what`, this means the name can no longer be found with `git grep` or `cscope`.

Granted, a function will always do something non-function-like (e.g. treat arguments of different types generically). Otherwise it should just be a C function. But the macro should aim to be function-like, unless there is a good reason not to.

But in this case, the macro does not seem an improvement.

I also don't understand, does this fix an actual bug, or does it just convert the code to a macro? The commit message should be clearer. If this is just about shortening code by putting it into a macro, then I am not convinced about the benefits of the result.

lib/xfrm/sp.c
```diff
@@ -675,21 +659,12 @@ int xfrmnl_sp_parse(struct nlmsghdr *n, struct xfrmnl_sp **result)
 				goto errout;
 			}
 
-			if (tmpl->family == AF_INET)
-				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a4, sizeof (tmpl->id.daddr.a4));
-			else
-				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a6, sizeof (tmpl->id.daddr.a6));
-			xfrmnl_user_tmpl_set_daddr (sputmpl, addr);
+			xfrm_tmpl_to_xfrmnl_tmpl(sputmpl, tmpl);
```
thom311, line 662: if the point is to avoid leaks, then maybe use the cleanup-attribute:
```diff
diff --git i/lib/xfrm/sp.c w/lib/xfrm/sp.c
index cf106517ba31..d7ad82a5cb65 100644
--- i/lib/xfrm/sp.c
+++ w/lib/xfrm/sp.c
@@ -658,44 +658,46 @@ int xfrmnl_sp_parse(struct nlmsghdr *n, struct xfrmnl_sp **result)
 		struct xfrm_userpolicy_type* up = nla_data(tb[XFRMA_POLICY_TYPE]);
 		memcpy ((void *)&sp->uptype, (void *)up, sizeof (struct xfrm_userpolicy_type));
 		sp->ce_mask     |= XFRM_SP_ATTR_POLTYPE;
 	}
 
 	if (tb[XFRMA_TMPL]) {
 		struct xfrm_user_tmpl*      tmpl = nla_data(tb[XFRMA_TMPL]);
 		struct xfrmnl_user_tmpl*    sputmpl;
 		uint32_t                    i;
 		uint32_t                    num_tmpls = nla_len(tb[XFRMA_TMPL]) / sizeof (*tmpl);
-		struct  nl_addr*            addr;
 
 		for (i = 0; (i < num_tmpls) && (tmpl); i ++, tmpl++)
 		{
+			_nl_auto_nl_addr struct nl_addr *addr1 = NULL;
+			_nl_auto_nl_addr struct nl_addr *addr2 = NULL;
+
 			if ((sputmpl = xfrmnl_user_tmpl_alloc ()) == NULL)
 			{
 				err = -NLE_NOMEM;
 				goto errout;
 			}
 
 			if (tmpl->family == AF_INET)
-				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a4, sizeof (tmpl->id.daddr.a4));
+				addr1 = nl_addr_build(tmpl->family, &tmpl->id.daddr.a4, sizeof (tmpl->id.daddr.a4));
 			else
-				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a6, sizeof (tmpl->id.daddr.a6));
-			xfrmnl_user_tmpl_set_daddr (sputmpl, addr);
+				addr1 = nl_addr_build(tmpl->family, &tmpl->id.daddr.a6, sizeof (tmpl->id.daddr.a6));
+			xfrmnl_user_tmpl_set_daddr (sputmpl, addr1);
 			xfrmnl_user_tmpl_set_spi (sputmpl, ntohl(tmpl->id.spi));
 			xfrmnl_user_tmpl_set_proto (sputmpl, tmpl->id.proto);
 			xfrmnl_user_tmpl_set_family (sputmpl, tmpl->family);
 
 			if (tmpl->family == AF_INET)
-				addr = nl_addr_build(tmpl->family, &tmpl->saddr.a4, sizeof (tmpl->saddr.a4));
+				addr2 = nl_addr_build(tmpl->family, &tmpl->saddr.a4, sizeof (tmpl->saddr.a4));
 			else
-				addr = nl_addr_build(tmpl->family, &tmpl->saddr.a6, sizeof (tmpl->saddr.a6));
-			xfrmnl_user_tmpl_set_saddr (sputmpl, addr);
+				addr2 = nl_addr_build(tmpl->family, &tmpl->saddr.a6, sizeof (tmpl->saddr.a6));
+			xfrmnl_user_tmpl_set_saddr (sputmpl, addr2);
 
 			xfrmnl_user_tmpl_set_reqid (sputmpl, tmpl->reqid);
 			xfrmnl_user_tmpl_set_mode (sputmpl, tmpl->mode);
 			xfrmnl_user_tmpl_set_share (sputmpl, tmpl->share);
 			xfrmnl_user_tmpl_set_optional (sputmpl, tmpl->optional);
 			xfrmnl_user_tmpl_set_aalgos (sputmpl, tmpl->aalgos);
 			xfrmnl_user_tmpl_set_ealgos (sputmpl, tmpl->ealgos);
 			xfrmnl_user_tmpl_set_calgos (sputmpl, tmpl->calgos);
 			xfrmnl_sp_add_usertemplate (sp, sputmpl);
 
```

Also, if
```C
			if (tmpl->family == AF_INET)
				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a4, sizeof (tmpl->id.daddr.a4));
			else
				addr = nl_addr_build(tmpl->family, &tmpl->id.daddr.a6, sizeof (tmpl->id.daddr.a6));
```
is too verbose, then add a helper function for that repeated code.

spellingmistake: It seems that in commit 3f4f1dda a `nl_put_addr` was just fine. I'll modify the patch accordingly.

PR #368, opened 2023-11-28
Implement can_device_stats

jobfeikens:


libnl-route-3.sym
```diff
@@ -1312,4 +1312,5 @@ global:
 	rtnl_link_bond_set_hashing_type;
 	rtnl_link_bond_set_miimon;
 	rtnl_link_bond_set_min_links;
+	rtnl_link_can_get_device_stats;
```
thom311, line 1315: please squash this commit in the previous.

the previous commit is incomplete/wrong, without this follow-up fix. No need to merge the change in two parts.

include/netlink/route/link/can.h
```diff
@@ -60,6 +60,9 @@ extern int rtnl_link_can_get_data_bittiming(struct rtnl_link *,
 extern int rtnl_link_can_set_data_bittiming(struct rtnl_link *,
                                             const struct can_bittiming *);
 
+int rtnl_link_can_get_device_stats(struct rtnl_link *,
```
thom311: please prefix the commit message with "route/can: " (and use lower-case for starting the sentence).

Could you also write one sentence what this is about?

lib/route/link/can.c
```diff
@@ -166,6 +168,11 @@ static int can_parse(struct rtnl_link *link, struct nlattr *data,
 		ci->ci_mask |= CAN_HAS_DATA_BITTIMING_CONST;
 	}
 
+	if (xstats) {
+		nla_memcpy(&ci->ci_device_stats, xstats, sizeof(ci->ci_device_stats));
```
thom311, line 172: `nla_memcpy()` will take care to copy only `_NL_MIN(count, nla_len(src))` bytes.

But it seems wrong to accept a truncated struct. I think it should be `if (xstats && nla_len(xstats) >= sizeof(ci->ci_devices_stats))`.

lib/route/link/can.c
```diff
@@ -864,6 +893,30 @@ int rtnl_link_can_set_data_bittiming(struct rtnl_link *link,
...
+	struct can_info *ci = link->l_info;
+
+	IS_CAN_LINK_ASSERT(link);
+	if (!device_stats)
+		return -NLE_INVAL;
+
+	if (ci->ci_mask & CAN_HAS_DEVICE_STATS)
+		*device_stats = ci->ci_device_stats;
+	else
+		return -NLE_AGAIN;
```
thom311, line 915: `return -NLE_MISSING_ATTR` ?

lib/route/link/can.c
```diff
@@ -352,6 +376,10 @@ static int can_put_attrs(struct nl_msg *msg, struct rtnl_link *link)
 
 	nla_nest_end(msg, data);
 
+	if (ci->ci_mask & CAN_HAS_DEVICE_STATS)
+		NLA_PUT(msg, IFLA_INFO_XSTATS, sizeof(ci->ci_device_stats),
+			&ci->ci_device_stats);
```
thom311, line 381: is this useful? There is no public API to set the device-stats, at most you could have gotten it from kernel. But then sending it back, seems wrong.

thom311 (changes requested):
thanks for your efforts.

lib/route/link/can.c
jobfeikens: NLE_MISSING_ATTR seems like a better error here. However,  all other getters in this file also return NLE_AGAIN when the attribute is absent. Should the other getters also be changed to NLE_MISSING_ATTR?

jobfeikens:
I applied the requested changes, please let me know what you think!

PR #372, opened 2024-01-08
route: Add extra support for bridge vlan and dynamic bridge table.

Cordell-O:


thom311:
Thank you for the patch!

There are some issues in CI (compiler warnings). Are you able to address those?

you can reformat the files by running `./tools/clang-format.sh`. That requires that you have a suitable clang-format version installed. Alternatively, you can use `./tools/clang-format-container.sh` (which requires a working podman and will build a Fedora container). If both is too cumbersome, you can also ignore the formatting issue, and I will reformat before merging.

lib/route/link/bridge.c
```diff
@@ -774,11 +917,38 @@ int rtnl_link_bridge_set_self(struct rtnl_link *link)
...
+ * @return -NLE_OPNOTSUP Link is not a bridge
+ */
+int rtnl_link_bridge_set_master(struct rtnl_link *link)
+{
+	struct bridge_data *bd = bridge_data(link);
+
+	IS_BRIDGE_LINK_ASSERT(link);
+
+	bd->b_master |= 1;
+	bd->b_self = 0;
```
thom311, line 945: if `s_master` and `s_self` are only set to 1, why the `|= 1`?

Anyway, can we make those two fields `bool` and use `true`/`false` ?


Also, `b_master` is only an internal field, and it's always `!b_self`. Are there future plans for this? Just drop the redundant field?

Maybe it would be better to support instead the `IFLA_BRIDGE_FLAGS`, and not trying to abstract individual flags (`BRIDGE_FLAGS_SELF`, `BRIDGE_FLAGS_MASTER`).

thom311 (changes requested):

Cordell-O:
Sorry for the slow response time, ive address all feedback, unsure about the test suite any feedback is appreciated.

lib/route/link/bridge.c
```diff
@@ -915,6 +1090,138 @@ uint16_t rtnl_link_bridge_str2hwmode(const char *name)
...
+
+	IS_BRIDGE_LINK_ASSERT(link);
+
+	vinfo = rtnl_link_bridge_get_port_vlan(link);
+
+	if (!vinfo)
+		return -NLE_NOATTR;
+
+	if (start > end || end >= RTNL_LINK_BRIDGE_VLAN_BITMAP_MAX)
+		return NLE_INVAL;
```
thom311, line 1137: `return -NLE_INVAL;`

lib/route/link/bridge.c
```diff
@@ -915,6 +1090,138 @@ uint16_t rtnl_link_bridge_str2hwmode(const char *name)
...
+
+	if (!vinfo)
+		return -NLE_NOATTR;
+
+	if (start > end || end >= RTNL_LINK_BRIDGE_VLAN_BITMAP_MAX)
+		return -NLE_INVAL;
+
+	for (uint16_t i = start; i <= end; i++)
+	{
+		if (i != vinfo->pvid)
```
thom311, line 1178: I think the doctext should mention, that pvid is ignored and not unset.

lib/route/link/bridge.c
```diff
@@ -915,6 +1090,138 @@ uint16_t rtnl_link_bridge_str2hwmode(const char *name)
...
+	struct rtnl_link_bridge_vlan * vinfo;
+
+	IS_BRIDGE_LINK_ASSERT(link);
+
+	vinfo = rtnl_link_bridge_get_port_vlan(link);
+
+	if (!vinfo)
+		return -NLE_NOATTR;
+
+	if (pvid >= RTNL_LINK_BRIDGE_VLAN_BITMAP_MAX)
```
thom311, line 1212: it seems, the function should also support to unset the pvid. E.g. accept `pvid=4096` to unset the value. And that should be documented.

lib/route/link/bridge.c
```diff
@@ -915,6 +1090,138 @@ uint16_t rtnl_link_bridge_str2hwmode(const char *name)
...
+		return -NLE_INVAL;
+
+	vinfo->pvid = pvid;
+
+	set_bit(pvid,vinfo->vlan_bitmap);
+	if (untagged)
+		set_bit(pvid,vinfo->untagged_bitmap);
```
thom311, line 1219: setting the pvid sets also these bits.

However, when another pvid was set previously, it does not unset it.

I think, `rtnl_link_bridge_set_port_vlan_map_range()`, `rtnl_link_bridge_unset_port_vlan_map_range()` and `rtnl_link_bridge_set_port_vlan_pvid()` should not have this odd interaction. Which is also inconsistent. For example, if you set `rtnl_link_bridge_set_port_vlan_map_range()` with `untagged=false`, then `rtnl_link_bridge_set_port_vlan_map_range(tagged=true)` will overwrite it.

I think these functions should be completely separate. Add in the documentation of `rtnl_link_bridge_set_port_vlan_pvid()`, that the user most likely also want to call `rtnl_link_bridge_set_port_vlan_map_range()` for that pvid.

lib/route/link/bridge.c
```diff
@@ -915,6 +1090,138 @@ uint16_t rtnl_link_bridge_str2hwmode(const char *name)
...
+		return -NLE_NOATTR;
+
+	if (start > end || end >= RTNL_LINK_BRIDGE_VLAN_BITMAP_MAX)
+		return NLE_INVAL;
+
+	for (uint16_t i = start; i <= end; i++)
+	{
+		set_bit(i,vinfo->vlan_bitmap);
+		if (untagged)
+			set_bit(i,vinfo->untagged_bitmap);
```
thom311, line 1143: shouldn't there be an `else unset_bit()` ?

(and make it clearer in the doc, that the `untagged` argument is applied, both to set/unset the bit).

include/nl-priv-static-route/nl-priv-static-route.h
```diff
@@ -3,6 +3,10 @@
 #ifndef NETLINK_ROUTE_UTILS_PRIV_H_
 #define NETLINK_ROUTE_UTILS_PRIV_H_
 
+#include <netlink/route/link/bridge.h>
+
 extern const uint8_t *const _nltst_map_stat_id_from_IPSTATS_MIB_v2;
+extern int bridge_fill_vlan_info(struct nl_msg *msg,
```
thom311, line 9: please give code in our headers a "nl" prefix. In this case, `_nl_bridge_fill_vlan_info()`.

tests/check-direct.c
```diff
@@ -53,12 +58,134 @@ START_TEST(static_checks)
...
+				"Expected attribute IFLA_BRIDGE_VLAN_INFO %d",
+				IFLA_BRIDGE_VLAN_INFO);
+			vlan_attr = (struct bridge_vlan_info *)nla_data(a);
+			ck_assert_int_eq(vlan_attr->vid,
+					 expected_attr[i][attr_count].vid);
+			ck_assert_int_eq(vlan_attr->flags,
+					 expected_attr[i][attr_count].flags);
+			attr_count++;
+		}
+		nlmsg_free(msg);
```
thom311, line 177:
```C
for (...) {
    _nl_auto_nl_msg struct nl_msg *msg = nlmsg_alloc();
    int attr_count = 0;
    int rem;
```


and no explicit `nlmsg_free()`.

tests/check-direct.c
```diff
@@ -53,12 +58,134 @@ START_TEST(static_checks)
...
+}
+END_TEST
+
 static Suite *make_suite(void)
 {
 	Suite *suite = suite_create("Direct");
 	TCase *tc = tcase_create("Core");
 
 	tcase_add_test(tc, static_checks);
+	tcase_add_test(tc, vlan_attribute_check);
```
thom311, line 188: thank you for your effort, to add a test!

thom311:
lgtm. Merged to `main`.

Thank you for your contribution!!

PR #375, opened 2024-04-21
Unused symbol versioning

lch361:
I have detected this problem while I was using different linkers for this project.

In this version script file `libnl-3.sym`, a symbol `nl_debug_dp` is defined. However, when doing `./configure --disable-debug`, there is no `nl_debug_dp` variable around, so a symbol is undefined. If the linker is configured to deny unused versions, this is a build error. Building with `ld.bfd` usually succeeds, because its option `--undefined-version` is set by default, while building with `ld.lld` fails because its `--undefined-version` isn't set by default.

This patch fixes this, enabling `--undefined-version` flag for every linker for every build.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -421,7 +421,8 @@ lib_libnl_3_la_CPPFLAGS = \
 	$(lib_cppflags)
 lib_libnl_3_la_LDFLAGS = \
 	-version-info $(LT_CURRENT):$(LT_REVISION):$(LT_AGE) \
-	-Wl,--version-script=$(srcdir)/libnl-3.sym
+	-Wl,--version-script=$(srcdir)/libnl-3.sym \
+	-Wl,--undefined-version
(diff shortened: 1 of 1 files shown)


thom311:
thanks for the report and the patch.

But I think this is the wrong solution.

The linker error about bugs in the `.sym` files is very useful. The unit tests in github actions also enable that linker flag (regardless whether it's the default). Granted, they didn't catch the problem because they didn't build with `--enable-debug=no`. That is fixed now by commit 13ab0122fc38462d16b2aa920167220ca196f15d.

I guess, if the ABI heavily depends on configure options, then disabling the linker check may make some sense. But libnl doesn't do that (except for `nl_debug_dp`), and it's probably a bad idea anyway.

This should be solved now with merge commit 96ddcd999297d8697235e75417bda8a66e26bb0f.

PR #378, opened 2024-04-24
Allow to allocate cache manager with custom refill socket

ievenbach:
Cache managers use two sockets: one for cache refill operation,
and another one for notifications.

In order to simulate NETLINK events by reading data from files,
we need to be able to overwrite callbacks for both sockets.

This new function allows us to set up refill socket any way we want.
It does have requirement that the refill socket be blocking.

### DIFF
--- a/include/netlink/cache.h
+++ b/include/netlink/cache.h
@@ -140,10 +140,15 @@ extern struct nl_cache *	__nl_cache_mngt_require(const char *);
 
 struct nl_cache_mngr;
 
-#define NL_AUTO_PROVIDE		1
-#define NL_ALLOCATED_SOCK	2  /* For internal use only, do not use */
+#define NL_AUTO_PROVIDE		    1
+#define NL_ALLOCATED_SOCK	    2  /* For internal use only, do not use */
+#define NL_ALLOCATED_SYNC_SOCK	4  /* For internal use only, do not use */
 
 extern int			nl_cache_mngr_alloc(struct nl_sock *,
+						    int, int,
+						    struct nl_cache_mngr **);
+extern int			nl_cache_mngr_alloc_ex(struct nl_sock *,
+							struct nl_sock *,
 						    int, int,
 						    struct nl_cache_mngr **);
 extern int			nl_cache_mngr_add(struct nl_cache_mngr *,
--- a/lib/cache_mngr.c
+++ b/lib/cache_mngr.c
@@ -149,16 +149,45 @@ found:
 int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
 			struct nl_cache_mngr **result)
 {
+	return nl_cache_mngr_alloc_ex(sk, NULL, protocol, flags, result);
+}
+
+/**
+ * Allocate new cache manager, with custom callback on refill socket
+ * @arg sk		Netlink socket or NULL to auto allocate
+ * @arg sync_sk		Blocking Netlink socket for cache refills
+ * @arg protocol	Netlink protocol this manager is used for
+ * @arg flags		Flags (\c NL_AUTO_PROVIDE)
+ * @arg result		Result pointer
+ *
+ * Same as \f nl_cache_mngr_alloc, but sets custom refill socket
+ * Note: ownership of the sync_sk passes to the cache manager
+ */
+int nl_cache_mngr_alloc_ex(struct nl_sock *sk, struct nl_sock *sync_sk, int protocol, int flags,
+			struct nl_cache_mngr **result)
+{
 	struct nl_cache_mngr *mngr;
 	int err = -NLE_NOMEM;
 
 	/* Catch abuse of flags */
-	if (flags & NL_ALLOCATED_SOCK)
+	if (flags & (NL_ALLOCATED_SOCK|NL_ALLOCATED_SYNC_SOCK))
 		BUG();
 
 	mngr = calloc(1, sizeof(*mngr));
-	if (!mngr)
+	if (!mngr) {
 		return -NLE_NOMEM;
+	}
+
+	if(!sync_sk) {
+		if (!(sync_sk = nl_socket_alloc()))
+			goto errout;
+
+		flags |= NL_ALLOCATED_SOCK;
+	}
+	/* have to assign before calling nl_connect, so that it gets freed in case
+	 * of an nl_socket_allock error for sk
+	 */
+	mngr->cm_sync_sock = sync_sk;
 
 	if (!sk) {
 		if (!(sk = nl_socket_alloc()))
@@ -176,6 +205,10 @@ int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
 	if (!mngr->cm_assocs)
 		goto errout;
 
+	if ((err = nl_connect(sync_sk, protocol)) < 0) {
+		goto errout;
+	}
+
 	/* Required to receive async event notifications */
 	nl_socket_disable_seq_check(mngr->cm_sock);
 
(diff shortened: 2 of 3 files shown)


include/netlink/cache.h
```diff
@@ -146,6 +146,10 @@ struct nl_cache_mngr;
 extern int			nl_cache_mngr_alloc(struct nl_sock *,
 						    int, int,
 						    struct nl_cache_mngr **);
+extern int			nl_cache_mngr_alloc_cb(struct nl_sock *,
+						    int, int,
+						    struct nl_cache_mngr **,
+						    struct nl_cb *sync_cb);
```
thom311, line 152: new ABI must be listed in `libnl-3.sym` file.

thom311 (commented):
> In order to simulate NETLINK events by reading data from files,
we need to be able to overwrite both sockets.

Passing a `nl_cb` argument for the internal socket doesn't really overwrite the socket. Wouldn't it be better to have instead a `struct nl_sock *sync_sk` argument?

ievenbach:
> > In order to simulate NETLINK events by reading data from files,
> > we need to be able to overwrite both sockets.
> 
> Passing a `nl_cb` argument for the internal socket doesn't really overwrite the socket. Wouldn't it be better to have instead a `struct nl_sock *sync_sk` argument?

The intent is not to overwrite the socket. I want as much of the functionality as possible intact. Instead, I want to provide a callback that overwrites the send/recv.

Why we need a separate socket at all is a different question.

ievenbach:
> > > In order to simulate NETLINK events by reading data from files,
> > > we need to be able to overwrite both sockets.
> > 
> > 
> > Passing a `nl_cb` argument for the internal socket doesn't really overwrite the socket. Wouldn't it be better to have instead a `struct nl_sock *sync_sk` argument?
> 
> The intent is not to overwrite the socket. I want as much of the functionality as possible intact. Instead, I want to provide a callback that overwrites the send/recv.
> 
> Why we need a separate socket at all is a different question.

Bah. I should read my own commit messages better. Fixed the description.

thom311:
> Why we need a separate socket at all is a different question.

That is the real question. Having two sockets (one for async events and the other for sync fetches) unavoidably means that it's not possibly to bring the messages in a consistent state. This can only correctly work, by using one socket for everything, where all events (sync and async) are in order.

Anyway. That is out of scope for this PR and would be a large work. I don't think anybody is going to fix that.

include/netlink/cache.h
thom311: The new entry MUST be added to the bottom of the file. See the comment at the top of the file.

thom311:
> The intent is not to overwrite the socket. I want as much of the functionality as possible intact. Instead, I want to provide a callback that overwrites the send/recv.

I see.

Since this is quite a special method, I still think it would be better to just accept a full `struct nl_sock *sync_sock` argument. For this more elaborate use-case, it provides the most flexibility, as the user can prepare the socket in any way the want. It's only mildly more cumbersome for the caller. Since `struct nl_sock` is not ref-counted, the function necessarily will take ownership of what you pass in.

Also, note that the `result` argument is an out-argument. Let's keep that the last parameter of the function.

lib/cache_mngr.c
```diff
@@ -185,23 +220,13 @@ int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
...
 		goto errout;
 
-	/* Create and allocate socket for sync cache fills */
-	mngr->cm_sync_sock = nl_socket_alloc();
-	if (!mngr->cm_sync_sock) {
-		err = -NLE_NOMEM;
-		goto errout;
-	}
-	if ((err = nl_connect(mngr->cm_sync_sock, protocol)) < 0)
-		goto errout_free_sync_sock;
```
thom311, line 195: not how for `mngr->cm_sock`, we call connect and non-blocking above. That means, for the `sk` argument, the caller is *not* expected to do that. That may be a questionable choice, but it is probably better to handle his consistently. Meaning: leave the `nl_connect(cm_sync_sock)` call here, also for the externally provided `sk_blocking`.

lib/cache_mngr.c
```diff
@@ -148,6 +148,33 @@ static int event_input(struct nl_msg *msg, void *arg)
...
+ * @arg sk		Netlink socket or NULL to auto allocate
+ * @arg sk_blocking		Blocking Netlink socket for cache refills
+ * @arg protocol	Netlink protocol this manager is used for
+ * @arg flags		Flags (\c NL_AUTO_PROVIDE)
+ * @arg result		Result pointer
+ *
+ * Same as \f nl_cache_mngr_alloc, but sets custom refill socket
+ * Note: ownership of the sk_blocking passes to the cache manager
+ */
+int nl_cache_mngr_alloc_ex(struct nl_sock *sk, struct nl_sock *sk_blocking, int protocol, int flags,
```
thom311, line 176: `sk_blocking` is internally assigned to `cm_sync_sock`. While it's unclear which is the best name, they probably should be consistent. I think the parameter should be called `sync_sk`.

lib/cache_mngr.c
```diff
@@ -156,10 +183,18 @@ int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
...
 
+	if(!sk_blocking) {
+		BUG();
+	}
+
 	mngr = calloc(1, sizeof(*mngr));
-	if (!mngr)
+	if (!mngr) {
+		// No manager yet, to free it in case of error
+		nl_socket_free(sk_blocking);
```
thom311, line 193: this is inconsistent too what happens with `sk`. On failure, the sockets are left alone. The caller would need to handle them in that case. That is also missing in `nl_cache_mngr_alloc()` (but that code will go away).

lib/cache_mngr.c
```diff
@@ -148,6 +148,33 @@ static int event_input(struct nl_msg *msg, void *arg)
...
+	int err;
+
+	if (!sk_blocking) {
+		return -NLE_NOMEM;
+	}
+	if ((err = nl_connect(sk_blocking, protocol)) < 0) {
+			nl_socket_free(sk_blocking);
+			return err;
+	}
+	return nl_cache_mngr_alloc_ex(sk, sk_blocking, protocol, flags, result);
```
thom311, line 162: as said, the logic of `nl_connect()` should not be on the caller.

Hence, most of this code is still in `nl_cache_mngr_alloc_ex()`.

Allow NULL as `sk_blocking` (`sync_sk`), and then just do
```C
int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
			struct nl_cache_mngr **result)
{
    return nl_cache_mngr_alloc_ex(sk, NULL, protocol, flags, result);
}
```

lib/cache_mngr.c
```diff
@@ -185,23 +220,13 @@ int nl_cache_mngr_alloc(struct nl_sock *sk, int protocol, int flags,
...
-		goto errout_free_sync_sock;
 
 	NL_DBG(1, "Allocated cache manager %p, protocol %d, %d caches\n",
 	       mngr, protocol, mngr->cm_nassocs);
 
 	*result = mngr;
 	return 0;
 
-errout_free_sync_sock:
-	nl_socket_free(mngr->cm_sync_sock);
```
thom311, line 204: like `sk`, the new parameter `sync_sk`/`sk_blocking` should be optional/nullable. If the user did not specify a blocking socket, in case of error the internally allocated one will still be needed to released.

You need to right logic, so that all internally allocated data is correctly freed on every error path. Use `_nl_auto_nl_socket` instead of explicit free for that.

thom311 (changes requested):

thom311:
Note also NL_ALLOCATED_SOCK, where ownership is only assumed for internally allocated socket. You would need something similar for sync-sock.

lib/cache_mngr.c
ievenbach: That was how I made it originally. However I noticed that presents a problem: we allow passing in two sockets, but cannot differentiate between errors caused by either. Which means we would need extra steps to be taken, to disown them internally in case of error, so that caller could release these.
I just found it easier to be inconsistent with socket assumptions.
I can fix the assumption about the second socket as well to be consistent.

ievenbach:
> Note also NL_ALLOCATED_SOCK, where ownership is only assumed for internally allocated socket. You would need something similar for sync-sock.

I rewrote the patch to follow this pattern.

thom311:
merged, with two follow-up changes.

Thank you!

ievenbach:
> merged, with two follow-up changes.
> 
> Thank you!

Thanks!

PR #379, opened 2024-04-24
Fix ubsan complaint about incorrect left-shift in generate_local_port()

ievenbach:
The value of "n" used to calculate get the unique port within application is 10 bit.
This change applies corresponding mask to n, prior to left-shifting it.

### DIFF
--- a/lib/socket.c
+++ b/lib/socket.c
@@ -84,7 +84,7 @@ static NL_RW_LOCK(port_map_lock);
 static uint32_t generate_local_port(void)
 {
 	int i, j, m;
-	uint16_t n;
+	uint32_t n;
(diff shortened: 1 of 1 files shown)


thom311:
I am not sure that this is right. `n & 0x3ff` is a positive signed integer of at most 0x3ff. If you shift that signed value by 22 bits to the left, then the value can become negative, which is again undefined behavior. I think, this would require at least `n & 0x3ffu`.

I think the problem here is that shifting a u16 value by 22 is undefined behavior. If so, a better fix is `((uint32_t) n) << 22)`.

Does that also avoid the ubsan issue?

thom311:
merged. Thank you!!

PR #380, opened 2024-04-24
Fix some errors, when using with C++

ievenbach:
1. "restrict" keyword does not exist in C++
2. Use explicit cast from malloc to parameter type used by _nl_inet_ntop

Note: this isn't enough to use pure C++. E.g. some macros use typeof() operator,
which is GCC extension.

### DIFF
--- a/include/base/nl-base-utils.h
+++ b/include/base/nl-base-utils.h
@@ -258,8 +258,7 @@ static inline bool _nl_memeq(const void *s1, const void *s2, size_t len)
 	return _nl_memcmp(s1, s2, len) == 0;
 }
 
-static inline void *_nl_memcpy(void *restrict dest, const void *restrict src,
-			       size_t n)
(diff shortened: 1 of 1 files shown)


ievenbach:
Run clang-format

thom311:
> The library itself is being built with C (clang). I had to put it into GCC mode, which is annoying, but OK.
The C++ part is the app I'm using.

I still don't understand what problem this patch solves.

The file you modify is an internal file. It's only used for building libnl library itself.

On the other hand, throughout the code base there are plenty implicit `void*` casts. So even with those two changes, you still cannot compile the library with C++.

What exactly are you doing? And how does this patch make that work?

ievenbach:
Looks like I just didn't realize it's private header. I got everything to compile when I turned on GNU extensions for typeof(), and plain forgot about it.

PR #384, opened 2024-05-14
route: ignore flags and weight of nexthops when doing a strict route comparison

KanjiMonster:
When a route is created while the interface has no link, we get a
notification with the route and the nexthop having the flag `LINKDOWN`.

If the interface later gets a link, we do not get a route notification
about it, so the route and nexthop stay at `LINKDOWN` in the libnl cache.

If the route then gets removed again, the to be removed route will not
have the `LINKDOWN` flag anymore, which then can break comparison of the
nexthop(s).

So add a new helper for ignoring FLAGS and WEIGHT for nexthops, and use it
everywhere we want to tell if it's the same nexthop, but not necessarily
in the same state.

For loose comparisons, we still need to use the old function, as else e.g.
route lookup will require specifying exact nexthops, and just filtering for
all routes on a certain interfaces would break.

### DIFF
--- a/lib/route/nexthop.c
+++ b/lib/route/nexthop.c
@@ -136,6 +136,23 @@ int rtnl_route_nh_compare(struct rtnl_nexthop *a, struct rtnl_nexthop *b,
 	return diff;
 }
 
+/**
+ * Check if the fixed attributes of two nexthops are identical, and may
+ * only differ in flags or weight.
+ *
+ * @arg a		a nexthop
+ * @arg b		another nexthop
+ *
+ * @return true if both nexthop have equal attributes, otherwise false.
+ */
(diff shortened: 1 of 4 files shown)


lib/route/nexthop.c
```diff
@@ -136,6 +136,14 @@ int rtnl_route_nh_compare(struct rtnl_nexthop *a, struct rtnl_nexthop *b,
 	return diff;
 }
 
+int rtnl_route_nh_identical(struct rtnl_nexthop *a, struct rtnl_nexthop *b)
```
thom311, line 139: could you please add a short doc-text for public API?

KanjiMonster: Sure thing, will add one

thom311:
merged. thank you!!!

PR #386, opened 2024-05-15
route: expose nexthop id attribute

KanjiMonster:
Routes may reference a nexthop (group) via the new nexthop API by its ID, so add accessors for setting and getting it.

Referencing a nexthop is mutually exclusive to specifiying nexthops in the route, so make sure we do not do that when creating netlink messages (which may exist both, since netlink messages from the kernel contain both unless 'nexthop_compat_mode' is disabled).

```
$ ip -6 r
2001:db8:3::/64 nhid 20 metric 1024 pref medium
	nexthop via 2001:db8:1::2 dev v0 weight 1
	nexthop via 2001:db8:2::2 dev v1 weight 1
```

Before:

```
$ nl-route-list
inet6 2001:db8:3::/64 table main type unicast via 2001:db8:1::2 dev v0 via 2001:db8:2::2 dev v1
```

After:

```
$ nl-route-list
inet6 2001:db8:3::/64 table main type unicast nhid 20 via 2001:db8:1::2 dev v0 via 2001:db8:2::2 dev v1
```

### DIFF
--- a/lib/route/route_obj.c
+++ b/lib/route/route_obj.c
@@ -65,6 +65,7 @@ struct rtnl_route {
 	uint32_t rt_metrics[RTAX_MAX];
 	uint32_t rt_metrics_mask;
 	uint32_t rt_nr_nh;
+	uint32_t rt_nhid;
 	struct nl_addr *rt_pref_src;
 	struct nl_list_head rt_nexthops;
 	struct rtnl_rtcacheinfo rt_cacheinfo;
@@ -90,6 +91,7 @@ struct rtnl_route {
 #define ROUTE_ATTR_REALMS    0x010000
 #define ROUTE_ATTR_CACHEINFO 0x020000
 #define ROUTE_ATTR_TTL_PROPAGATE 0x040000
+#define ROUTE_ATTR_NHID      0x080000
 /** @endcond */
 
 static void route_constructor(struct nl_object *c)
@@ -194,6 +196,9 @@ static void route_dump_line(struct nl_object *a, struct nl_dump_params *p)
 	if (r->ce_mask & ROUTE_ATTR_TOS && r->rt_tos != 0)
 		nl_dump(p, "tos %#x ", r->rt_tos);
 
+	if (r->ce_mask & ROUTE_ATTR_NHID)
+		nl_dump(p, "nhid %u ", r->rt_nhid);
+
 	if (r->ce_mask & ROUTE_ATTR_MULTIPATH) {
 		struct rtnl_nexthop *nh;
 
@@ -283,6 +288,9 @@ static void route_dump_details(struct nl_object *a, struct nl_dump_params *p)
 		nl_dump(p, " ttl-propagate %s",
 			r->rt_ttl_propagate ? "enabled" : "disabled");
 	}
+
+	if (r->ce_mask & ROUTE_ATTR_NHID)
+		nl_dump(p, "nhid %u ", r->rt_nhid);
 
 	nl_dump(p, "\n");
 
@@ -414,6 +422,7 @@ static uint64_t route_compare(struct nl_object *_a, struct nl_object *_b,
 		      nl_addr_cmp(a->rt_pref_src, b->rt_pref_src));
 	diff |= _DIFF(ROUTE_ATTR_TTL_PROPAGATE,
 		      a->rt_ttl_propagate != b->rt_ttl_propagate);
+	diff |= _DIFF(ROUTE_ATTR_NHID, a->rt_nhid != b->rt_nhid);
 
 	if (flags & LOOSE_COMPARISON) {
 		nl_list_for_each_entry(nh_b, &b->rt_nexthops, rtnh_list) {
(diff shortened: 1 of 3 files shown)


KanjiMonster:
I didn't try to link the nexthop and route caches for now, to keep the changes simple.

lib/route/route_obj.c
```diff
@@ -284,6 +289,9 @@ static void route_dump_details(struct nl_object *a, struct nl_dump_params *p)
 			r->rt_ttl_propagate ? "enabled" : "disabled");
 	}
 
+	if (r->ce_mask & ROUTE_ATTR_NHID && r->rt_nhid != 0)
+		nl_dump(p, "nhid %d ", r->rt_nhid);
```
thom311, line 293: `%u`

that is appropriate for the type of `r->rt_nhid`, but it's also what iproute2 does.

lib/route/route_obj.c
```diff
@@ -284,6 +289,9 @@ static void route_dump_details(struct nl_object *a, struct nl_dump_params *p)
 			r->rt_ttl_propagate ? "enabled" : "disabled");
 	}
 
+	if (r->ce_mask & ROUTE_ATTR_NHID && r->rt_nhid != 0)
```
thom311, line 292: there needs to be consistency about what `rt_nhid=0` means.

Note that to kernel (`ip nexthop add id 0 dev eth0`), zero ID means to choose an unused ID.

Below, `rtnl_route_set_nhid()` accepts setting to 0, while also setting the `ROUTE_ATTR_NHID`.


What needs to be done, is that the code never checks for `r->rt_nhid != 0` directly. Instead, this is a ID-based next-hop if-and-only-if `ROUTE_ATTR_NHID` is set.

In this case, during dump we want to see when the nhid is set to zero and not hide that (important) fact.

lib/route/route_obj.c
```diff
@@ -1339,6 +1361,10 @@ int rtnl_route_parse(struct nlmsghdr *nlh, struct rtnl_route **result)
 			return err;
 	}
 
+	if (tb[RTA_NH_ID]) {
+		rtnl_route_set_nhid(route, nla_get_u32(tb[RTA_NH_ID]));
```
thom311, line 1365: I suspect, that kernel would never send RTA_NH_ID set to zero? Then this code is fine.

Otherwise, what does it mean, if kernel would send RTA_NH_ID set to zero? I think, to libnl that should mean that `ROUTE_ATTR_NHID` is *not* set, and we should skip calling `rtnl_route_set_nhid()`.

But as I suspect that can never happen (kernel wouldn't send this), it's fine as is.

lib/route/route_obj.c
```diff
@@ -1441,7 +1467,10 @@ int rtnl_route_build_msg(struct nl_msg *msg, struct rtnl_route *route)
 		nla_nest_end(msg, metrics);
 	}
 
-	if (rtnl_route_get_nnexthops(route) == 1) {
+	/* Nexthop specification and nexthop id are mutually exclusive */
+	if (route->rt_nhid > 0) {
```
thom311, line 1471: `if (mask & ROUTE_ATTR_NHID)`

Note that it is a questionable feature to let kernel choose the nexthop ID. However, if the user called `rtnl_route_set_nhid(route, 0)`, then that's what they should get.

thom311 (changes requested):
good patch! Minor commonts.

lib/route/route_obj.c
KanjiMonster: In the kernel RTA_NH_ID is stored in a (zeroed) struct, so 0 has here the meaning that there is no named nexthop, not that the kernel should choose/create one and assign an id.

All later checks are for `cfg->fc_nh_id > 0`. So for routes, RTA_NH_ID 0 has a slightly different meaning.

How about changing the setter to clear `ROUTE_ATTR_NHID` if being set to 0? Then it would mirror the meaning in the kernel, and we can use `ROUTE_ATTR_NHID` to mean it is set to a valid value. And you have a way of unsetting the value.

thom311: > How about changing the setter to clear ROUTE_ATTR_NHID if being set to 0? Then it would mirror the meaning in the kernel, and we can use ROUTE_ATTR_NHID to mean it is set to a valid value. And you have a way of unsetting the value.

That sounds best!

KanjiMonster: Done.

lib/route/route_obj.c
KanjiMonster: Fixed, and yes, %u sounds more appropriate.

thom311:
There is still a `if (r->ce_mask & ROUTE_ATTR_NHID && r->rt_nhid != 0)`. It should only check the flags (granted, the conditions are true always at the same time).

Rest lgtm.

KanjiMonster:
> There is still a `if (r->ce_mask & ROUTE_ATTR_NHID && r->rt_nhid != 0)`. It should only check the flags (granted, the conditions are true always at the same time).
> 
> Rest lgtm.

Oops, fixed.

thom311:
merged as 3e0806317b5116bca7a097a0cb9b7a9cbebc465a


Thank you!!

PR #389, opened 2024-05-29
route: treat routes with via nexthops as universe scoped as well

KanjiMonster:
`RTA_VIA` is a `RTA_GATEWAY` with added address family, so we should treat them them equivalent for guessing the scope.

### DIFF
--- a/lib/route/route_obj.c
+++ b/lib/route/route_obj.c
@@ -1026,7 +1026,7 @@ int rtnl_route_guess_scope(struct rtnl_route *route)
 		 * is not directly connected
 		 */
 		nl_list_for_each_entry(nh, &route->rt_nexthops, rtnh_list) {
-			if (nh->rtnh_gateway)
+			if (nh->rtnh_gateway || nh->rtnh_via)
 				return RT_SCOPE_UNIVERSE;
 		}
 	}


KanjiMonster:
> Error: file "include/base/nl-base-utils.h" has style issues.

I didn't even touch that file ...

thom311:
> > Error: file "include/base/nl-base-utils.h" has style issues.
> 
> I didn't even touch that file ...

this seems to be a change in behavior of clang-format in Fedora 40. That's a bit annoying.

The version shipped by current Fedora 40 is the correct version for formatting. The file needs reformatting, I will do that.

thom311:
merged as 326882017876bc22f9d16508a44a8ef98a53d220.

Thank you for the patch!!

thom311:
> > > Error: file "include/base/nl-base-utils.h" has style issues.
> > 
> > 
> > I didn't even touch that file ...
> 
> this seems to be a change in behavior of clang-format in Fedora 40. That's a bit annoying.
> 
> The version shipped by current Fedora 40 is the correct version for formatting. The file needs reformatting, I will do that.

if you use `./tools/clang-format-container.sh` for formatting (which may be a good idea, if you don't run Fedora 40), then you would need to regenerate the container. Just issue a `podman rmi localhost/libnl-code-format-f40`.

KanjiMonster:
> > > Error: file "include/base/nl-base-utils.h" has style issues.
> > 
> > 
> > I didn't even touch that file ...
> 
> this seems to be a change in behavior of clang-format in Fedora 40. That's a bit annoying.
> 
> The version shipped by current Fedora 40 is the correct version for formatting. The file needs reformatting, I will do that.

The weird/annyoing thing is that the ["original" run](https://github.com/thom311/libnl/actions/runs/9133239943/job/25116272437) ran fine with clang-format 18.1.1, but clang-format 18.1.6 now triggers this.

A version bump that minor shouldn't change behaviour ... .

> if you use ./tools/clang-format-container.sh for formatting (which may be a good idea, if you don't run Fedora 40), then you would need to regenerate the container.

Well, since the file I touched was part of the ignored files in `./tools/clang-format.sh`, I didn't think to even have to run clang-format ;-).

I did run a manual clang-format call on the file itself though, to make sure the code I touched is at least not introducing *more* issues.

thom311:
@KanjiMonster I think I would release libnl-3.10 soon. Do you have anything that still should go in and to wait for?

KanjiMonster:
I have nothing urgent, so feel free to release whenever. All open things I have open are either too small to be a blocker, or too large to get fixed/implemented soon.

We are using a heavily patched libnl anyway, so I'm no stranger to backporting things if I need them.

PR #391, opened 2024-06-26
route/link: add bonding parser/getters and lacp rate api

eguillaume-kalray:
Enables getting attributes from bonding interfaces.

### DIFF
--- a/include/netlink/route/link/bonding.h
+++ b/include/netlink/route/link/bonding.h
@@ -26,10 +26,19 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
 extern int	rtnl_link_bond_release(struct nl_sock *, struct rtnl_link *);
 
 extern void	rtnl_link_bond_set_mode(struct rtnl_link *link, uint8_t mode);
+extern uint8_t	rtnl_link_bond_get_mode(struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_activeslave(struct rtnl_link *link, int active_slave);
+extern int	rtnl_link_bond_get_activeslave(struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_hashing_type (struct rtnl_link *link, uint8_t type);
+extern uint8_t	rtnl_link_bond_get_hashing_type (struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_miimon (struct rtnl_link *link, uint32_t miimon);
+extern uint32_t	rtnl_link_bond_get_miimon (struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_min_links (struct rtnl_link *link, uint32_t min_links);
+extern uint32_t	rtnl_link_bond_get_min_links (struct rtnl_link *link);
 
 #ifdef __cplusplus
 }
--- a/include/netlink/route/link/bonding.h
+++ b/include/netlink/route/link/bonding.h
@@ -40,6 +40,9 @@ extern uint32_t	rtnl_link_bond_get_miimon (struct rtnl_link *link);
 extern void	rtnl_link_bond_set_min_links (struct rtnl_link *link, uint32_t min_links);
 extern uint32_t	rtnl_link_bond_get_min_links (struct rtnl_link *link);
 
+extern void	rtnl_link_bond_set_lacp_rate(struct rtnl_link *link, uint8_t lacp_rate);
+extern uint8_t	rtnl_link_bond_get_lacp_rate(struct rtnl_link *link);
+
 #ifdef __cplusplus
 }
 #endif
--- a/libnl-route-3.sym
+++ b/libnl-route-3.sym
@@ -1331,4 +1331,9 @@ global:
 	rtnl_route_get_nhid;
 	rtnl_route_nh_identical;
 	rtnl_route_set_nhid;
+	rtnl_link_bond_get_mode;
+	rtnl_link_bond_get_hashing_type;
+	rtnl_link_bond_get_activeslave;
+	rtnl_link_bond_get_miimon;
+	rtnl_link_bond_get_min_links;
 } libnl_3_9;
--- a/libnl-route-3.sym
+++ b/libnl-route-3.sym
@@ -1336,4 +1336,6 @@ global:
 	rtnl_link_bond_get_activeslave;
 	rtnl_link_bond_get_miimon;
 	rtnl_link_bond_get_min_links;
+	rtnl_link_bond_set_lacp_rate;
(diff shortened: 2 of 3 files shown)


include/netlink/route/link/bonding.h
```diff
@@ -26,10 +26,22 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
 extern int	rtnl_link_bond_release(struct nl_sock *, struct rtnl_link *);
 
 extern void	rtnl_link_bond_set_mode(struct rtnl_link *link, uint8_t mode);
+extern uint8_t	rtnl_link_bond_get_mode(struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_activeslave(struct rtnl_link *link, int active_slave);
+extern int	rtnl_link_bond_get_activeslave(struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_hashing_type (struct rtnl_link *link, uint8_t type);
+extern uint8_t	rtnl_link_bond_get_hashing_type (struct rtnl_link *link);
```
KanjiMonster, line 35: Maybe also fix the style here
```suggestion
extern int 	rtnl_link_bond_get_hashing_type(struct rtnl_link *link, uint8_t *type);
```

include/netlink/route/link/bonding.h
```diff
@@ -26,10 +26,22 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
...
+
 extern void	rtnl_link_bond_set_miimon (struct rtnl_link *link, uint32_t miimon);
+extern uint32_t	rtnl_link_bond_get_miimon (struct rtnl_link *link);
```
KanjiMonster, line 38: likewise
```suggestion
extern int	rtnl_link_bond_get_miimon(struct rtnl_link *link, uint32_t *miimon);
```

include/netlink/route/link/bonding.h
```diff
@@ -26,10 +26,22 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
 extern int	rtnl_link_bond_release(struct nl_sock *, struct rtnl_link *);
 
 extern void	rtnl_link_bond_set_mode(struct rtnl_link *link, uint8_t mode);
+extern uint8_t	rtnl_link_bond_get_mode(struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_activeslave(struct rtnl_link *link, int active_slave);
+extern int	rtnl_link_bond_get_activeslave(struct rtnl_link *link);
```
KanjiMonster, line 32: The usual way for these is to return an int, and store the value in a pointer passed as argument, so you can actually return error values if the value isn't set.

```suggestion
extern int	rtnl_link_bond_get_activeslave(struct rtnl_link *link, int *active_slave);
```

E.g. activeslave wouldn't exist for an LACP bond, so we should return an error here.

libnl-route-3.sym
```diff
@@ -1331,4 +1331,11 @@ global:
 	rtnl_route_get_nhid;
 	rtnl_route_nh_identical;
 	rtnl_route_set_nhid;
+	rtnl_link_bond_get_mode;
+	rtnl_link_bond_get_hashing_type;
+	rtnl_link_bond_get_activeslave;
+	rtnl_link_bond_get_miimon;
+	rtnl_link_bond_get_min_links;
+	rtnl_link_bond_set_lacp_rate;
+	rtnl_link_bond_get_lacp_rate;
```
KanjiMonster, line 1340: Alphabetical order?

include/netlink/route/link/bonding.h
```diff
@@ -26,10 +26,22 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
...
+
 extern void	rtnl_link_bond_set_hashing_type (struct rtnl_link *link, uint8_t type);
+extern uint8_t	rtnl_link_bond_get_hashing_type (struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_miimon (struct rtnl_link *link, uint32_t miimon);
+extern uint32_t	rtnl_link_bond_get_miimon (struct rtnl_link *link);
+
 extern void	rtnl_link_bond_set_min_links (struct rtnl_link *link, uint32_t min_links);
+extern uint32_t	rtnl_link_bond_get_min_links (struct rtnl_link *link);
```
KanjiMonster, line 41:
```suggestion
extern int	rtnl_link_bond_get_min_links(struct rtnl_link *link, uint32_t *min_links);
```

include/netlink/route/link/bonding.h
```diff
@@ -26,10 +26,22 @@ extern int	rtnl_link_bond_release_ifindex(struct nl_sock *, int);
...
+
+extern void	rtnl_link_bond_set_lacp_rate(struct rtnl_link *link, uint8_t lacp_rate);
+extern uint8_t	rtnl_link_bond_get_lacp_rate(struct rtnl_link *link);
```
KanjiMonster, line 44:
```suggestion
extern int	rtnl_link_bond_get_lacp_rate(struct rtnl_link *link, uint8_t *lacp_rate);
```

include/netlink/route/link/bonding.h
thom311: I also think so. Only returning `uint8_t` leaves no space to signal an error (or a missing attribute `-NLE_NOATTR`).

If the value itself is a `uint8_t`, then we could still do
```C
int rtnl_link_bond_get_hashing_type (struct rtnl_link *link);
```
where negative values are error codes and non-negative values are the `uint8_t` value.

For `uint32_t` types, that doesn't work. So it needs
```C
extern int rtnl_link_bond_get_miimon(struct rtnl_link *link, uint32_t *miimon);
```
to separately signal the error/NOATTR and the value.


I think `int rtnl_link_bond_get_hashing_type (struct rtnl_link *link);` is still less convenient to use, because you need
```C
    r = rtnl_link_bond_get_hashing_type(link);
    if (r >= 0)
         hashing_type = r;
    else {
          /* handle error */
    }
```
instead of
```C
     r = rtnl_link_bond_get_hashing_type(link, &hashing_type);
     if (r < 0) {
         /* handle error */
     }
```

So just have getters return an `int` (zero or negative error code) and the out value separately. Ensure that the out value is touched, if-and-only-if there is no error.

thom311 (changes requested):
(setting flag for "Request changes")

KanjiMonster:
To get this moving in this I decided to push my changes I had already prepared a while ago while I was waiting for the 3.10 release with https://github.com/thom311/libnl/pull/401.

PR #398, opened 2024-08-15
lib/attr: add nla functions for variable-length integers

Geywn:
The NLA_{UINT|SINT} attributes[1] are introduced for variable-length integers, whose length are 32bits or 64bits.

So add appropriate functions to get or put NLA_{UINT|SINT} attributes.

[1] https://lore.kernel.org/netdev/20231018213921.2694459-3-kuba@kernel.org/

### DIFF
--- a/include/netlink/attr.h
+++ b/include/netlink/attr.h
@@ -47,6 +47,9 @@ enum {
 	NLA_S16,
 	NLA_S32,
 	NLA_S64,
+	/* Skip unimplemented attr types  */
+	NLA_SINT = 20,
+	NLA_UINT,
 	__NLA_TYPE_MAX,
 };
 
--- a/include/netlink/attr.h
+++ b/include/netlink/attr.h
@@ -121,6 +121,10 @@ extern int64_t          nla_get_s64(const struct nlattr *);
 extern int              nla_put_s64(struct nl_msg *, int, int64_t);
 extern uint64_t		nla_get_u64(const struct nlattr *);
 extern int		nla_put_u64(struct nl_msg *, int, uint64_t);
+extern int64_t		nla_get_sint(const struct nlattr *);
+extern int              nla_put_sint(struct nl_msg *, int, int64_t);
+extern uint64_t		nla_get_uint(const struct nlattr *);
+extern int              nla_put_uint(struct nl_msg *, int, uint64_t);
 
 /* String attribute */
 extern char *		nla_get_string(const struct nlattr *);
--- a/lib/attr.c
+++ b/lib/attr.c
@@ -776,6 +776,70 @@ uint64_t nla_get_u64(const struct nlattr *nla)
 	return tmp;
 }
 
+/**
+ * Add variable-length signed integer attribute to netlink message.
+ * @arg msg             Netlink message.
+ * @arg attrtype        Attribute type.
+ * @arg value           Numeric value to store as payload.
+ *
+ * @see nla_put
+ * @return 0 on success or a negative error code.
+ */
+int nla_put_sint(struct nl_msg *msg, int attrtype, int64_t value)
+{
+	int64_t tmp64 = value;
+	int32_t tmp32 = value;
+
+	if (tmp64 == tmp32)
+		return nla_put_s32(msg, attrtype, tmp32);
+	return nla_put(msg, attrtype, sizeof(int64_t), &value);
+}
+
+/**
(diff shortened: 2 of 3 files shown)


lib/attr.c
```diff
@@ -774,6 +774,70 @@ uint64_t nla_get_u64(const struct nlattr *nla)
...
+ * @see nla_put
+ * @return 0 on success or a negative error code.
+ */
+int nla_put_uint(struct nl_msg *msg, int attrtype, uint64_t value)
+{
+	uint64_t tmp64 = value;
+	uint32_t tmp32 = value;
+
+	if (tmp64 == tmp32)
+		nla_put_u32(msg, attrtype, tmp32);
```
thom311, line 824: return.

include/netlink/attr.h
```diff
@@ -118,6 +118,10 @@ extern int64_t          nla_get_s64(const struct nlattr *);
 extern int              nla_put_s64(struct nl_msg *, int, int64_t);
 extern uint64_t		nla_get_u64(const struct nlattr *);
 extern int		nla_put_u64(struct nl_msg *, int, uint64_t);
+extern int64_t		nla_get_sint(const struct nlattr *);
+extern int              nla_put_sint(struct nl_msg *, int, int64_t);
+extern uint64_t		nla_get_uint(const struct nlattr *);
+extern int              nla_put_uint(struct nl_msg *, int, uint64_t);
```
thom311, line 124: doesn't this lack a `NLA_UINT` enum value (beside `NLA_U32`)?

Note that so far, the enum values (like `NLA_U32`) in `include/netlink/attr.h` mirror the ones from kernel's `include/net/netlink.h`. I think, you should (possibly in a separate commit) copy the enum over from the kernel header, so that they are the same numeric values. Even if it (probably?) would not matter if they have different values...

Maybe NLA types that are not implemented by libnl (e.g. `NLA_BITFIELD32`, `NLA_REJECT`) should be excluded, however, ensure that the numeric values of `NLA_UINT` is the same as the one in kernel.

Then also adjust things like `nla_attr_minlen`.

thom311:
thanks for the patch!

lib/attr.c
```diff
@@ -774,6 +774,70 @@ uint64_t nla_get_u64(const struct nlattr *nla)
...
+ * @see nla_put
+ * @return 0 on success or a negative error code.
+ */
+int nla_put_sint(struct nl_msg *msg, int attrtype, int64_t value)
+{
+	int64_t tmp64 = value;
+	int32_t tmp32 = value;
+
+	if (tmp64 == tmp32)
+		nla_put_s32(msg, attrtype, tmp32);
```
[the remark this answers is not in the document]
Geywn: Oops, apologize for my carelessness. I will fix it. Thanks!

lib/attr.c
Geywn: Yes, will fix it. Thanks!

include/netlink/attr.h
Geywn: Thanks for pointing out. Will do: 1) define NLA_SINT/UINT as values in kernel's include/net/netlink.h. 2) adjust nla_attr_minlen().

Geywn:
change log:
v2->v1:
1. add a new patch to introduce NLA_{SINT|UINT} enum value to `include/netlink/attr.h`, and adjust `nla_attr_minlen`.
2. fix the missing `return` in the 2nd patch.

lib/attr.c
```diff
@@ -776,6 +776,70 @@ uint64_t nla_get_u64(const struct nlattr *nla)
...
+	return nla_put(msg, attrtype, sizeof(uint64_t), &value);
+}
+
+/**
+ * Return payload of variable-length unsigned integer attribute
+ * @arg nla		variable-length unsigned integer attribute
+ *
+ * @return Payload as 64 bit integer.
+ */
+uint64_t nla_get_uint(const struct nlattr *nla)
```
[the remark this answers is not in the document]
Geywn: Sure, thanks!

Geywn:
change log:
v3->v2:
expose nla_{get|put}_{s|u}int ABI in libnl-3.sym.

v2->v1:
add a new patch to introduce NLA_{SINT|UINT} enum value to include/netlink/attr.h, and adjust nla_attr_minlen.
fix the missing return in the 2nd patch.

thom311:
merged. Thank you!!

PR #401, opened 2024-08-27
bond link improvements

KanjiMonster:
Add various missing pieces around bond link support:

* implement comparing bond link objects
* implement cloning bond link objects
* add getters for bond attributes
* implement parsing link info of bond link objects

### DIFF
--- a/lib/route/link/bonding.c
+++ b/lib/route/link/bonding.c
@@ -54,6 +54,59 @@ static int bond_info_alloc(struct rtnl_link *link)
 	return 0;
 }
 
+static struct nla_policy bonding_nl_policy[IFLA_BOND_MAX + 1] = {
+	[IFLA_BOND_MODE]             = { .type = NLA_U8 },
+	[IFLA_BOND_ACTIVE_SLAVE]     = { .type = NLA_U32 },
+	[IFLA_BOND_MIIMON]	     = { .type = NLA_U32 },
+	[IFLA_BOND_XMIT_HASH_POLICY] = { .type = NLA_U8 },
+	[IFLA_BOND_MIN_LINKS]	     = { .type = NLA_U32 },
+};
+
+static int bond_info_parse(struct rtnl_link *link, struct nlattr *data,
+			   struct nlattr *xstats)
+{
+	struct nlattr *tb[IFLA_BOND_MAX + 1];
+	struct bond_info *bn;
+	int err;
+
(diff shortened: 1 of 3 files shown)


lib/route/link/bonding.c
```diff
@@ -59,6 +59,28 @@ static void bond_info_free(struct rtnl_link *link)
 	_nl_clear_free(&link->l_info);
 }
 
+static int bond_info_clone(struct rtnl_link *dst, struct rtnl_link *src)
+{
+	struct bond_info *bond_dst, *bond_src = src->l_info;
+	int err;
+
+	if (!bond_src)
+		BUG();
```
thom311, line 121: Let's not use `BUG()`.

Either do `assert(bond_src)` or `_nl_assert(bond_src)`.

The difference between those two assertion macros is that the former is usually enabled by default (even production builds tend NOT to define `NDEBUG`). The latter is usually disabled (unless you build with `-DNL_MORE_ASSERTS=level`).

`_nl_assert()` is for expensive checks or things where we are especially confident that the condition won't be broken in future refactoring.

So probably `assert()` is better here. Both is fine.

thom311:
I only gave it a quick look, but looks pretty good!! Will check more.

lib/route/link/bonding.c
thom311: Actually, in this case, I think `_nl_assert()` is better.

The `_nl_assert()` makes it clear to the reviewer what we expect, but it's a NOP on most builds. However, it will still be easy to discover/debug the bug even if the assertion is not active, because there will be a crash immediately afterwards.

KanjiMonster: There are multiple link type handlers using BUG() or BUG_ON(), so I was just following that. Anyway, changed it to `_nl_assert()`.

lib/route/link/bonding.c
```diff
@@ -59,6 +59,28 @@ static void bond_info_free(struct rtnl_link *link)
...
+	err = bond_info_alloc(dst);
+	if (err)
+		return err;
+
+	bond_dst = dst->l_info;
+
+	if (!bond_dst)
+		BUG();
+
+	memcpy(bond_dst, bond_src, sizeof(struct bond_info));
```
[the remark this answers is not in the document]
KanjiMonster: Updated.

thom311:
very nice. Thank you!

PR #402, opened 2024-08-30
add auto cleanup helper for nl_object and use it

KanjiMonster:
Add a auto cleanup helper for `nl_object`, and use it where appropriate for cache and xfrm.

### DIFF
--- a/lib/cache.c
+++ b/lib/cache.c
@@ -714,17 +714,14 @@ static int __cache_pickup(struct nl_sock *sk, struct nl_cache *cache,
 static int pickup_checkdup_cb(struct nl_object *c, struct nl_parser_param *p)
 {
 	struct nl_cache *cache = (struct nl_cache *)p->pp_arg;
-	struct nl_object *old;
+	_nl_auto_nl_object struct nl_object *old = NULL;
 
 	old = nl_cache_search(cache, c);
 	if (old) {
-		if (nl_object_update(old, c) == 0) {
-			nl_object_put(old);
+		if (nl_object_update(old, c) == 0)
 			return 0;
-		}
 
 		nl_cache_remove(old);
-		nl_object_put(old);
 	}
 
 	return nl_cache_add(cache, c);
@@ -788,8 +785,8 @@ static int cache_include(struct nl_cache *cache, struct nl_object *obj,
 			 struct nl_msgtype *type, change_func_t cb,
 			 change_func_v2_t cb_v2, void *data)
 {
-	struct nl_object *old;
-	struct nl_object *clone = NULL;
+	_nl_auto_nl_object struct nl_object *old = NULL;
+	_nl_auto_nl_object struct nl_object *clone = NULL;
 	uint64_t diff = 0;
 
 	switch (type->mt_act) {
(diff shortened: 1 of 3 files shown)


lib/cache.c
```diff
@@ -714,17 +714,14 @@ static int __cache_pickup(struct nl_sock *sk, struct nl_cache *cache,
 static int pickup_checkdup_cb(struct nl_object *c, struct nl_parser_param *p)
 {
 	struct nl_cache *cache = (struct nl_cache *)p->pp_arg;
-	struct nl_object *old;
+	_nl_auto_nl_object struct nl_object *old = NULL;
 
 	old = nl_cache_search(cache, c);
```
KanjiMonster, line 719: I only set this to `NULL` because it triggers a false positive in the CI gcc:

```
In file included from ./include/nl-default.h:8,
                 from lib/cache.c:46:
In function '_nl_auto_nl_object_fcn',
    inlined from 'pickup_checkdup_cb' at lib/cache.c:717:39:
./include/base/nl-base-utils.h:754:20: error: 'old' may be used uninitialized [-Werror=maybe-uninitialized]
  754 |                 if (*v)                                  \
      |                    ^
./include/nl-aux-core/nl-core.h:56:1: note: in expansion of macro '_NL_AUTO_DEFINE_FCN_TYPED0'
   56 | _NL_AUTO_DEFINE_FCN_TYPED0(struct nl_object *, _nl_auto_nl_object_fcn,
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~
lib/cache.c: In function 'pickup_checkdup_cb':
lib/cache.c:717:46: note: 'old' was declared here
  717 |         _nl_auto_nl_object struct nl_object *old;
      |                                              ^~~
```

Go home GCC, you are drunk.

Changing both to one line

```
_nl_auto_nl_object struct nl_object *old = nl_cache_search(cache, c);
```

is also accepted by gcc.

I tried to replicate it locally, but no version I tried complained (11,12,14).

thom311: Yes, we must always initialize auto-variables that have a `cleanup` attribute. This is the way.

Even in cases like here, where we might deduce that the function never returns before initializing the variable.

I think that is related to `-fexceptions`, where an exception might reach the cleanup callback without initializing the variable (of course, most C code would not correctly handle exception at arbitrary places, so it's unclear to me what this is good for).

---

I think `_nl_auto_nl_object struct nl_object *old = NULL;` is slightly nicer, because otherwise the line gets so long. Also, initializing to `nl_cache_search(cache, c)` is not trivial, so it maybe shouldn't be cramped in one line.

KanjiMonster: > This is the way.

This is the way.

> I think that is related to `-fexceptions`, where an exception might reach the cleanup callback without initializing the variable (of course, most C code would not correctly handle exception at arbitrary places, so it's unclear to me what this is good for).

That at least makes a bit of sense, although mostly coming from kernel it is a foreign concept for me. There are no exceptions in kernel space.

> I think `_nl_auto_nl_object struct nl_object *old = NULL;` is slightly nicer, because otherwise the line gets so long. Also, initializing to `nl_cache_search(cache, c)` is not trivial, so it maybe shouldn't be cramped in one line.

I agree there, which is why I did so :-)

thom311:
merged as d48da13239f5.

Thank you for the patch.

IMO this again shows just how useful the cleanup attribute is. While in theory everybody can reason about the proper places where to call free/unref/put, in practice it is a very error prone effort. Instead, we should be explicit about ownership transfer with the trifecta `_nl_steal_pointer()`, `_nl_clear_pointer()` and `_nl_auto*`. Also, the features that gcc/clang provide on top of standard C, make it worth to depend on some of them.

PR #404, opened 2024-09-10
lib/route/link: add support for getting permanent mac address of link.

precla:
Couldn't find any function that gets an interface's permanent mac address (`LINK_ATTR_PERMANENT_ADDR`).

I hope that i did all, or at least most of it, the proper way. My guess is that i forgot to add it (`l_paddr`) on some places...

AFAIK: permanent address can not be changed

Waiting for feedback.

edit: pull request 404 :)

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -73,6 +73,7 @@
 #define LINK_ATTR_GSO_MAX_SEGS		((uint64_t) 1 << 37)
 #define LINK_ATTR_GSO_MAX_SIZE		((uint64_t) 1 << 38)
 #define LINK_ATTR_LINKINFO_SLAVE_KIND	((uint64_t) 1 << 39)
+#define LINK_ATTR_PERMANENT_ADDR	((uint64_t) 1 << 40)
 
 static struct nl_cache_ops rtnl_link_ops;
 static struct nl_object_ops link_obj_ops;
@@ -271,6 +272,7 @@ static void link_free_data(struct nl_object *c)
 
 		nl_addr_put(link->l_addr);
 		nl_addr_put(link->l_bcast);
+		nl_addr_put(link->l_paddr);
 
 		free(link->l_ifalias);
 		free(link->l_info_kind);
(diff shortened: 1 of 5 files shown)


libnl-route-3.sym
```diff
@@ -1331,4 +1331,10 @@ global:
 	rtnl_route_get_nhid;
 	rtnl_route_nh_identical;
 	rtnl_route_set_nhid;
+	rtnl_link_get_perm_addr;
```
thom311, line 1334: The unit tests fail, see the github actions.
I think it's due to this line (I suspect you can drop this line).

precla: indeed, some leftover while testing. removed. CI is passing now

libnl-route-3.sym
```diff
@@ -1331,7 +1331,6 @@ global:
 	rtnl_route_get_nhid;
 	rtnl_route_nh_identical;
 	rtnl_route_set_nhid;
-	rtnl_link_get_perm_addr;
```
thom311: This commit fixes a problem if the previous commit, within the same PR.

In this case, could you please instead squash the commits, so that there is only commit. The important part is not that there is only one commit, but that each commit is self-contained as correct (as far as we know).

Also, could you please rebase to latest `main`? There will be a conflict, but it should be easy to resolve.

lib/route/link.c
```diff
@@ -1565,6 +1585,9 @@ int rtnl_link_fill_info(struct nl_msg *msg, struct rtnl_link *link)
 	if (link->ce_mask & LINK_ATTR_NS_PID)
 		NLA_PUT_U32(msg, IFLA_NET_NS_PID, link->l_ns_pid);
 
+	if (link->ce_mask & LINK_ATTR_PERMANENT_ADDR)
+		NLA_PUT_ADDR(msg, IFLA_PERM_ADDRESS, link->l_paddr);
```
thom311, line 1589: does this make sense? As there is no `rtnl_link_set_perm_addr()`, is this still useful?

Maybe add a `rtnl_link_set_perm_addr()`, or just remove these two lines. Or why not?

thom311 (changes requested):

libnl-route-3.sym
precla: done

lib/route/link.c
precla: doesn't make sense and is removed now

thom311:
merged as acf572b585361587740b7c80ed33b3df2706ece9.

Thank you!

PR #405, opened 2024-09-10
lib/route: add rtnl_neigh ext flags support

ronand-atl:
This PR adds support for extended neighbor flags which are sent as an attribute.

A use case for this feature is being able to detect if an FDB entry has the locked flag set.

### DIFF
--- a/include/netlink/route/neighbour.h
+++ b/include/netlink/route/neighbour.h
@@ -36,6 +36,9 @@ extern int	rtnl_neigh_str2state(const char *);
 extern char *	rtnl_neigh_flags2str(int, char *, size_t);
 extern int	rtnl_neigh_str2flag(const char *);
 
+extern char *	rtnl_neigh_extflags2str(uint32_t, char *, size_t);
+extern uint32_t	rtnl_neigh_str2extflag(const char *);
+
 extern int	rtnl_neigh_add(struct nl_sock *, struct rtnl_neigh *, int);
 extern int	rtnl_neigh_build_add_request(struct rtnl_neigh *, int,
 					     struct nl_msg **);
@@ -54,6 +57,13 @@ extern void			rtnl_neigh_set_flags(struct rtnl_neigh *,
 extern void			rtnl_neigh_unset_flags(struct rtnl_neigh *,
 						       unsigned int);
 extern unsigned int		rtnl_neigh_get_flags(struct rtnl_neigh *);
+
+extern void			rtnl_neigh_set_ext_flags(struct rtnl_neigh *,
+							 uint32_t);
+extern void			rtnl_neigh_unset_ext_flags(struct rtnl_neigh *,
+							   uint32_t);
+extern int			rtnl_neigh_get_ext_flags(struct rtnl_neigh *,
+							 uint32_t *);
 
 extern void			rtnl_neigh_set_ifindex(struct rtnl_neigh *,
 						       int);
--- a/lib/route/neigh.c
+++ b/lib/route/neigh.c
@@ -171,6 +171,7 @@ struct rtnl_neigh {
 	uint32_t n_ifindex;
 	uint16_t n_state;
 	uint8_t n_flags;
+	uint32_t n_ext_flags;
 	uint8_t n_type;
 	struct nl_addr *n_lladdr;
 	struct nl_addr *n_dst;
@@ -179,6 +180,7 @@ struct rtnl_neigh {
 	struct rtnl_ncacheinfo n_cacheinfo;
 	uint32_t n_state_mask;
 	uint32_t n_flag_mask;
+	uint32_t n_ext_flag_mask;
 	uint32_t n_master;
 	uint16_t n_vlan;
 };
@@ -191,10 +193,11 @@ struct rtnl_neigh {
 #define NEIGH_ATTR_IFINDEX      0x20
 #define NEIGH_ATTR_FAMILY       0x40
 #define NEIGH_ATTR_TYPE         0x80
-#define NEIGH_ATTR_PROBES       0x100
-#define NEIGH_ATTR_MASTER       0x200
-#define NEIGH_ATTR_VLAN         0x400
-#define NEIGH_ATTR_NHID         0x800
+#define NEIGH_ATTR_PROBES       0x0100
+#define NEIGH_ATTR_MASTER       0x0200
+#define NEIGH_ATTR_VLAN         0x0400
+#define NEIGH_ATTR_NHID         0x0800
+#define NEIGH_ATTR_EXT_FLAGS    0x1000
 
 static struct nl_cache_ops rtnl_neigh_ops;
 static struct nl_object_ops neigh_obj_ops;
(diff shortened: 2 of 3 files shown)


thom311:
hm, while internally kernel merges the 8 bit flags with the ext-flags
```C
    neigh_flags_ext = pn->flags >> NTF_EXT_SHIFT;
    neigh_flags     = pn->flags & NTF_OLD_MASK;
```
on the netlink API these are basically independent flags (for example, `NTF_EXT_MANAGED` is 0x1, and not 0x100). That seem an unfortunate decision, because those could just as well be the same flags (with the exception, that on netlink, flags larger than 8 bit are expressed in a new attribute).

Anyway.  Since in the kernel API this looks like separate flags, it is also right that they are treated this way in the libnl API. This approach requires us to add 5 new public API functions, when otherwise the existing API was already suitable to express 32bit flags.

include/netlink/route/neighbour.h
```diff
@@ -55,6 +58,12 @@ extern void			rtnl_neigh_unset_flags(struct rtnl_neigh *,
 						       unsigned int);
 extern unsigned int		rtnl_neigh_get_flags(struct rtnl_neigh *);
 
+extern void			rtnl_neigh_set_ext_flags(struct rtnl_neigh *,
+							 uint32_t);
+extern void			rtnl_neigh_unset_ext_flags(struct rtnl_neigh *,
+							   uint32_t);
+extern uint32_t			rtnl_neigh_get_ext_flags(struct rtnl_neigh *);
```
thom311, line 65: I think this should be
```C
extern int rtnl_neigh_get_ext_flags(struct rtnl_neigh *neigh, uint32_t *flags);
```

(and return `-NLE_NOATTR` if the flags are not present).


That would allow to see, whether kernel sent those flags.

Indeed, kernel does not send `NEIGH_ATTR_EXT_FLAGS` if they are all zero. That seems unfortunate, because then a user cannot see whether the attribute is just not supported or whether it happens to be zero. That also lessens the use of knowing in `rtnl_neigh_get_ext_flags()` whether the flags are actually set.

Still, I think this should be done. Even if it's inconsistent with `rtnl_neigh_get_flags()`. But for `rtnl_neigh_get_flags()` the flags come from `nm->ndm_flags`, so it would always be present even with older kernel versions.

lib/route/neigh.c
```diff
@@ -762,6 +777,9 @@ static int build_neigh_msg(struct rtnl_neigh *tmpl, int cmd, int flags,
 	if (tmpl->ce_mask & NEIGH_ATTR_NHID)
 		NLA_PUT_U32(msg, NDA_NH_ID, tmpl->n_nhid);
 
+	if (tmpl->ce_mask & NEIGH_ATTR_EXT_FLAGS)
+		NLA_PUT_U32(msg, NDA_FLAGS_EXT, tmpl->n_ext_flags);
```
thom311, line 781: kernel does:
```C
#define NTF_EXT_MASK      (NTF_EXT_MANAGED)
```
```C
[NDA_FLAGS_EXT]          = NLA_POLICY_MASK(NLA_U32, NTF_EXT_MASK),
```

Note that `NTF_EXT_LOCKED` is not part of the mask. That means, if we sent kernel a `NDA_FLAGS_EXT` attribute with  `NTF_EXT_LOCKED` flag set, it will be rejected.

This means, if we receive a neighbor instance from kernel, modify it slightly, and send it back, then it will be rejected.

As it never makes sense to send kernel such a flag (at least not today, until this restriction is lifted), it should be filtered out on the client side.

Also, if we don't have any flags to set, we maybe should not send the attribute. Older kernel versions would not know this attribute (though they may or may not ignore unknown netlink attributes to be forward compatible).


In summary, how about:
```C
    if (tmpl->ce_mask & NEIGH_ATTR_EXT_FLAGS) {
        uint32_t f = tmpl->n_ext_flags & (~((uint32_t) NTF_EXT_LOCKED));

        /* We cannot set NTF_EXT_LOCKED, filter it out.
         * Also, don't send the attribute if it's zero anyway, to be backward
         * compatible with older kernel. */
        if (f != 0)
           NLA_PUT_U32(msg, NDA_FLAGS_EXT, f);
    }
```

thom311 (changes requested):

lib/route/neigh.c
ronand-atl: > Also, if we don't have any flags to set, we maybe should not send the attribute. Older kernel versions would not know this attribute (though they may or may not ignore unknown netlink attributes to be forward compatible).

The code already has this behavior. The ext flags will only be sent to the kernel if `NEIGH_ATTR_EXT_FLAGS` is set in the `ce_mask`. This will only be set if the user has set or unset an ext flag themselves or the kernel sent a message with the `NDA_FLAGS_EXT` attribute.

I've added a commit to unset the locked flag in the request, and only send the attribute if the flags value is nonzero.

Should we also not allow the user to set the locked flag? That would require changing the function signature of the setter, unless a separate variable is stored which represents the user's desired setting for the locked flag (which is only used if set). In that case the kernel would report the error instead of libnl. These are just some ideas. Maybe that's overcomplicating things and the user should just not expect setting that flag to do anything (but not error either).

include/netlink/route/neighbour.h
ronand-atl: I can agree with this, although I do see the merits in keeping it consistent with `get_flags` since the absence of the attribute doesn't seem to allow the user to make any meaningful conclusion, and they would need to do more error checking.

> Indeed, kernel does not send NEIGH_ATTR_EXT_FLAGS if they are all zero

It seems the kernel may send ext flags even if they are all zero if dumping FDB entries (see `fdb_fill_info`).

I've added a commit to make the function error if the flags aren't present.

lib/route/neigh.c
thom311: > > Also, if we don't have any flags to set, we maybe should not send the attribute.
>
> The code already has this behavior. The ext flags will only be sent to the kernel if NEIGH_ATTR_EXT_FLAGS is set in the ce_mask.

Right. With "don't have any flags set", I meant that the flags are explicitly set to zero (with NEIGH_ATTR_EXT_FLAGS). Anyway.

> Should we also not allow the user to set the locked flag?

I think `rtnl_neigh_set_ext_flags()` should allow setting the flag (just as it does now). 
- It might make sense to record the LOCKED flag in a object that the user creates. 
- it's clear that libnl cannot possibly send this to kernel, so requiring the user to filter out the flag is not helpful.
- we set the flag when parsing it, so a user should also be able to create a similar object with the flag set.

lib/route/neigh.c
```diff
@@ -777,8 +777,15 @@ static int build_neigh_msg(struct rtnl_neigh *tmpl, int cmd, int flags,
...
-	if (tmpl->ce_mask & NEIGH_ATTR_EXT_FLAGS)
-		NLA_PUT_U32(msg, NDA_FLAGS_EXT, tmpl->n_ext_flags);
+	if (tmpl->ce_mask & NEIGH_ATTR_EXT_FLAGS) {
+		/* The kernel does not allow setting the locked flag from
+		 * userspace, so unset it in the request. */
+		uint32_t ext_flags = tmpl->n_ext_flags &
+				     ~(uint32_t)NTF_EXT_LOCKED;
+
+		if (ext_flags)
+			NLA_PUT_U32(msg, NDA_FLAGS_EXT, tmpl->n_ext_flags);
```
thom311, line 787: `NLA_PUT_U32(msg, NDA_FLAGS_EXT, ext_flags);`

Also, could you please squash the commits together? The follow-up commits merely fix the first commit, just rebase (force-push) the pull request.

ronand-atl: Oops, missed that, good catch.

And yeah I intended to squash once everything was figured out. I'll do that now.

ronand-atl:
Just fixed that small issue and squashed the commits. Let me know if there's any other changes you want me to make, otherwise it should be good to merge :)

thom311:
lgtm. Merged as 014c33a6e55b88630e334e25da0a23e0a62f4b4d

Thank you!

PR #406, opened 2024-09-13
Extend bridge info support

ronand-atl:
This PR adds a few more supported attributes in route/link/bridge_info.

Support for the following bridge attributes is added:
- IFLA_BR_STP_STATE
- IFLA_BR_MCAST_ROUTER
- IFLA_BR_MCAST_SNOOPING

Support for IFLA_BR_MULTI_BOOLOPT is added, so the following boolopts can be set:
- BR_BOOLOPT_NO_LL_LEARN
- BR_BOOLOPT_MCAST_VLAN_SNOOPING
- BR_BOOLOPT_MST_ENABLE

Additionally, I've added a few missing getter functions for already supported attributes.

### DIFF
--- a/include/netlink/route/link/bridge_info.h
+++ b/include/netlink/route/link/bridge_info.h
@@ -37,12 +37,21 @@ extern void rtnl_link_bridge_set_vlan_stats_enabled(struct rtnl_link *link,
 						    uint8_t vlan_stats_enabled);
 extern int rtnl_link_bridge_get_vlan_stats_enabled(struct rtnl_link *link,
 						   uint8_t *vlan_stats_enabled);
+
 extern void rtnl_link_bridge_set_nf_call_iptables(struct rtnl_link *link,
 						  uint8_t call_enabled);
+extern int rtnl_link_bridge_get_nf_call_iptables(struct rtnl_link *link,
+						 uint8_t *call_enabled);
+
 extern void rtnl_link_bridge_set_nf_call_ip6tables(struct rtnl_link *link,
 						   uint8_t call_enabled);
+extern int rtnl_link_bridge_get_nf_call_ip6tables(struct rtnl_link *link,
+						  uint8_t *call_enabled);
+
 extern void rtnl_link_bridge_set_nf_call_arptables(struct rtnl_link *link,
 						   uint8_t call_enabled);
+extern int rtnl_link_bridge_get_nf_call_arptables(struct rtnl_link *link,
+						  uint8_t *call_enabled);
 
 #ifdef __cplusplus
 }
--- a/include/netlink/route/link/bridge_info.h
+++ b/include/netlink/route/link/bridge_info.h
@@ -68,6 +68,11 @@ extern void rtnl_link_bridge_set_mcast_snooping(struct rtnl_link *link,
 extern int rtnl_link_bridge_get_mcast_snooping(struct rtnl_link *link,
 					       uint8_t *value);
 
+extern int rtnl_link_bridge_set_boolopt(struct rtnl_link *link, int opt,
+					int value);
+
+extern int rtnl_link_bridge_get_boolopt(struct rtnl_link *link, int opt);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/netlink/route/link/bridge_info.h
+++ b/include/netlink/route/link/bridge_info.h
@@ -53,6 +53,21 @@ extern void rtnl_link_bridge_set_nf_call_arptables(struct rtnl_link *link,
 extern int rtnl_link_bridge_get_nf_call_arptables(struct rtnl_link *link,
 						  uint8_t *call_enabled);
 
+extern void rtnl_link_bridge_set_stp_state(struct rtnl_link *link,
+					   uint32_t stp_state);
+extern int rtnl_link_bridge_get_stp_state(struct rtnl_link *link,
+					  uint32_t *stp_state);
+
+extern void rtnl_link_bridge_set_mcast_router(struct rtnl_link *link,
+					      uint8_t type);
+extern int rtnl_link_bridge_get_mcast_router(struct rtnl_link *link,
+					     uint8_t *type);
+
+extern void rtnl_link_bridge_set_mcast_snooping(struct rtnl_link *link,
+						uint8_t value);
+extern int rtnl_link_bridge_get_mcast_snooping(struct rtnl_link *link,
+					       uint8_t *value);
+
 #ifdef __cplusplus
 }
 #endif
--- a/lib/route/link/bridge_info.c
+++ b/lib/route/link/bridge_info.c
@@ -13,6 +13,8 @@
 
 #include "nl-default.h"
 
+#include <linux/if_bridge.h>
+
 #include <netlink/route/link/bridge_info.h>
 
 #include "nl-route.h"
@@ -29,6 +31,7 @@
 #define BRIDGE_ATTR_STP_STATE (1 << 8)
 #define BRIDGE_ATTR_MCAST_ROUTER (1 << 9)
 #define BRIDGE_ATTR_MCAST_SNOOPING (1 << 10)
+#define BRIDGE_ATTR_BOOLOPT (1 << 11)
 
 struct bridge_info {
 	uint32_t ce_mask; /* to support attr macros */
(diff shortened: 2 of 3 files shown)


libnl-route-3.sym
```diff
@@ -1340,5 +1340,9 @@ global:
 	rtnl_link_bond_get_miimon;
 	rtnl_link_bond_get_min_links;
 	rtnl_link_bond_get_mode;
+	rtnl_link_bridge_get_nf_call_arptables;
+	rtnl_link_bridge_get_nf_call_ip6tables;
+	rtnl_link_bridge_get_nf_call_iptables;
+	rtnl_link_bridge_set_vlan_default_pvid;
```
thom311, line 1358: Optimally, this bugfix would be a separate commit. Could you be bothered to split it out? :)

include/netlink/route/link/bridge_info.h
```diff
@@ -6,6 +6,7 @@
 #ifndef NETLINK_LINK_BRIDGE_INFO_H_
 #define NETLINK_LINK_BRIDGE_INFO_H_
 
+#include <stdbool.h>
```
thom311, line 9: `<stdbool.h>` is a C99 header. So far we don't use that in public headers.

However, we already use `<stdint.h>` (also C99), so we can assume that all users use C99 and will also have <stdbool.h>.

One concern is that `<stdbool.h>` redefines `bool`/`true`/`false`. So if a user had their own such identifiers, this change will break them (well, at least, if they use this particular header -- which probably means that nobody is going to be affected). However, this also creates a precedence, so soon we might use `<stdbool.h>` in more prominent headers, where it might break some users.

Unclear what to do about that. Maybe just use an `int`?

include/netlink/route/link/bridge_info.h
```diff
@@ -68,6 +69,19 @@ extern void rtnl_link_bridge_set_mcast_snooping(struct rtnl_link *link,
 extern int rtnl_link_bridge_get_mcast_snooping(struct rtnl_link *link,
 					       uint8_t *value);
 
+extern void rtnl_link_bridge_set_no_linklocal_learn(struct rtnl_link *link,
+						    bool value);
+extern int rtnl_link_bridge_get_no_linklocal_learn(struct rtnl_link *link,
+						   bool *value);
```
thom311, line 75: In this case, we can return the boolean value and an error value at the same time.

In earlier places, where we get a `uint32_t`, a `int` cannot encode a (negative) error value and the return value. So we had the output argument. I think for getters where successful values fit into `[0, INT_MAX]`, we can fit them both together. It's more convenient to use.

Consistency is a concern here, however, we are already not quite consistent (sometimes for good reasons). So, it seems more useful to me to use the API that makes the most sense for the type. Meaning, a getter for a boolean can squeeze the error and the boolean value in an `int`.

lib/route/link/bridge_info.c
```diff
@@ -239,6 +256,55 @@ static struct rtnl_link_info_ops bridge_info_ops = {
...
+
+	mask = 1ul << opt;
+
+	if (value)
+		bi->b_boolopts.optval |= mask;
+	else
+		bi->b_boolopts.optval &= ~mask;
+
+	bi->b_boolopts.optmask |= mask;
+	bi->ce_mask |= BRIDGE_ATTR_BOOLOPT;
```
thom311, line 278: There is no API to unset `bi->ce_mask & BRIDGE_ATTR_BOOLOPT` or `bi->b_boolopts.optmask & mask`.

That's a common lack in many setters, so it's probably fine. 

If we make `value` an `int`, we could take "-1" to unset `bi->b_boolopts.optmask`. Not sure that is useful though. It's also slightly confusing. So maybe not.

Fine-as-is, unless you'd like to change it.

lib/route/link/bridge_info.c
```diff
@@ -779,6 +845,99 @@ int rtnl_link_bridge_get_mcast_snooping(struct rtnl_link *link, uint8_t *value)
...
+ * @arg link	Link object of type bridge
+ * @arg value	Output argument.
+ *
+ * @see rtnl_link_bridge_set_no_linklocal_learn()
+ *
+ * @return Zero on success, otherwise a negative error code.
+ * @retval -NLE_NOATTR
+ * @retval -NLE_INVAL
+ */
+int rtnl_link_bridge_get_no_linklocal_learn(struct rtnl_link *link, bool *value)
```
thom311, line 873: I really think that instead of adding all those `rtnl_link_bridge_get_no_linklocal_learn()` accessors, just expose `bridge_info_set_boolopt()` as public API.

Note that enum values like `BR_BOOLOPT_NO_LL_LEARN` are already public API (in kernel headers). Just let the user include `if_bridge.h` and set the flags themselves. The benefit is, that we support also future enum values.

Also, no need for the check `if (opt < 0 || opt >= BR_BOOLOPT_MAX)`, if the user sets a value that is not supported then kernel will tell us. On the other hand, we must check that opt is < 32, so that the bitshift is not undefined behavior.

thom311 (changes requested):

libnl-route-3.sym
ronand-atl: Sure

include/netlink/route/link/bridge_info.h
ronand-atl: Personally I think that using an int over a bool for new code would probably be a step backwards. However, since your suggestion lines up nicely with your other suggestion, I will change it to use an int.

lib/route/link/bridge_info.c
ronand-atl: I also noticed the common lack of a way to unset values. I'm not sure of the usefulness of that though, since you could just not set the value in the first place. Maybe it would be useful for getting an object, modifying it a bit, and sending it back as a change request which matches a different object. That seems like an odd use case though. I think I'll leave it as is for now. If there is a desire in the future for ways to unset values, maybe a `reset` or `clear` version of the functions could be added.

lib/route/link/bridge_info.c
ronand-atl: I thought that might be brought up. I initially did it that way because I thought it might be nice to have a higher level API that abstracts the boolopts detail.

I've made the boolopts setter and getter public and removed the individual functions for now.

include/netlink/route/link/bridge_info.h
```diff
@@ -8,6 +8,7 @@
 
 #include <netlink/netlink.h>
 #include <netlink/route/link.h>
+#include <linux/if_bridge.h>
```
ronand-atl, line 11: This probably won't fly. Are there any other solutions other than just using an int instead of the enum?

include/netlink/route/link/bridge_info.h
ronand-atl: Not sure I'm sold on the lack of consistency argument but I'll change it anyway

include/netlink/route/link/bridge_info.h
thom311: Why do you say it won't fly?

I didn't previously realize that `BR_BOOLOPT_NO_LL_LEARN` is an enum. Often enough, such identifiers are just int (literals) in kernel header.

I don't think we should/need to use `enum br_boolopt_id`. We also avoid enums in our public API, because their size depends on compiler flags like `-fshort-enums` (not sure that's a very convincing argument).

I think just use a plain `int id`. Then also don't include `<linux/if_bridge.h>`. Let the user include that to get the enum values they need.

ronand-atl: > Why do you say it won't fly?

Probably because we would be including our copy of the kernel header which would inflate the header size. Also the included kernel header may be older or newer than the one on the user's system, which could be undesirable.

Changing to use an int instead.

thom311: >  It would use the kernel header that the user has on their system, not the one included in libnl.

Right. The one in libnl under `include/linux-private/` is only to build libnl itself. Those headers are not for the users of the library.

> That means you can't rely on the kernel header file defining a symbol that is used in the libnl header file.

That probably would be fine. We rely on kernel headers to be stable and not break API. In this case, `enum br_boolopt_id` was added in linux v5.0 (March 2019), so if we use the symbol, we would add a dependency on 5.0+ (which is old enough to be OK).

However, I still think using a plain "int" is preferable. Users who want to use this API should include `<linux/if_bridge.h>` themselves and use enums like `BR_BOOLOPT_NO_LL_LEARN`.

thom311:
merged.

Very nice, thank you!

PR #409, opened 2024-10-07
lib/route: add support for bridge msti

ronand-atl:
This PR adds the ability to set MSTP port states on bridge ports. There a high level API to set up vlan/msti associations yet, so the user would still have to do that part manually until #407 is been finished.

### DIFF
--- a/lib/route/link/bridge.c
+++ b/lib/route/link/bridge.c
@@ -14,11 +14,13 @@
 #include "nl-default.h"
 
 #include <linux/if_bridge.h>
+#include <linux/rtnetlink.h>
 
 #include <netlink/netlink.h>
 #include <netlink/attr.h>
 #include <netlink/route/rtnl.h>
 #include <netlink/route/link/bridge.h>
+#include <netlink/list.h>
 
 #include "nl-route.h"
 #include "link-api.h"
@@ -28,28 +30,36 @@
 #define VLAN_VID_MASK           0x0fff /* VLAN Identifier */
 
 /** @cond SKIP */
-#define BRIDGE_ATTR_PORT_STATE		(1 << 0)
-#define BRIDGE_ATTR_PRIORITY		(1 << 1)
-#define BRIDGE_ATTR_COST		(1 << 2)
-#define BRIDGE_ATTR_FLAGS		(1 << 3)
-#define BRIDGE_ATTR_PORT_VLAN           (1 << 4)
-#define BRIDGE_ATTR_HWMODE		(1 << 5)
-#define BRIDGE_ATTR_CONFIG_MODE		(1 << 6)
+#define BRIDGE_ATTR_PORT_STATE		(1UL << 0)
+#define BRIDGE_ATTR_PRIORITY		(1UL << 1)
+#define BRIDGE_ATTR_COST		(1UL << 2)
+#define BRIDGE_ATTR_FLAGS		(1UL << 3)
+#define BRIDGE_ATTR_PORT_VLAN           (1UL << 4)
+#define BRIDGE_ATTR_HWMODE		(1UL << 5)
+#define BRIDGE_ATTR_CONFIG_MODE		(1UL << 6)
+#define BRIDGE_ATTR_MST			(1UL << 7)
 
-#define PRIV_FLAG_NEW_ATTRS		(1 << 0)
+#define PRIV_FLAG_NEW_ATTRS		(1UL << 0)
 
 struct bridge_data
 {
-	uint8_t			b_port_state;
-	uint8_t			b_priv_flags; /* internal flags */
-	uint16_t		b_hwmode;
-	uint16_t		b_priority;
-	uint16_t		b_config_mode;
-	uint32_t		b_cost;
-	uint32_t		b_flags;
-	uint32_t		b_flags_mask;
-	uint32_t                ce_mask; /* HACK to support attr macros */
-	struct rtnl_link_bridge_vlan vlan_info;
+	uint8_t				b_port_state;
+	uint8_t				b_priv_flags; /* internal flags */
+	uint16_t			b_hwmode;
+	uint16_t			b_priority;
+	uint16_t			b_config_mode;
+	uint32_t			b_cost;
+	uint32_t			b_flags;
+	uint32_t			b_flags_mask;
+	uint32_t			ce_mask; /* HACK to support attr macros */
+	struct rtnl_link_bridge_vlan	vlan_info;
+	struct nl_list_head		mst_list;
+};
+
+struct mst_state_entry {
+	struct nl_list_head list_node;
+	uint16_t msti; /* unique within a list */
+	uint8_t state;
 };
 
 static void set_bit(unsigned nr, uint32_t *addr)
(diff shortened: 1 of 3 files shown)


lib/route/link/bridge.c
```diff
@@ -410,6 +577,39 @@ int _nl_bridge_fill_vlan_info(struct nl_msg *msg, struct rtnl_link_bridge_vlan *
...
+		nla_nest_end(msg, entry_attr);
+	}
+
+	nla_nest_end(msg, attr);
+
+	return 0;
+
+nla_put_failure:
+	nla_nest_cancel(msg, entry_attr);
+	nla_nest_cancel(msg, attr);
```
thom311, line 609: there are two explicit `goto nla_put_failure`. Those should not both cancel the nested attributes...

lib/route/link/bridge.c
```diff
@@ -675,6 +910,20 @@ static int bridge_compare(struct rtnl_link *_a, struct rtnl_link *_b,
...
+	({ \
+		uint64_t val = (uint64_t)(state) << 16 | (uint64_t)(msti + 1); \
+		val ^= val << 13; val ^= val >> 7; val ^= val << 17; val; })
+
+	nl_list_for_each_entry(entry, &a->mst_list, list_node) {
+		hash ^= _MST_HASH(entry->msti, entry->state);
+	}
+	nl_list_for_each_entry(entry, &b->mst_list, list_node) {
+		hash ^= _MST_HASH(entry->msti, entry->state);
+	}
```
thom311, line 923: This doesn't seem right. With `^` we indeed do some hashing of the values, but we want a equality implementation that bails out if there are any differences

lib/route/link/bridge.c
```diff
@@ -1376,6 +1625,186 @@ struct rtnl_link_bridge_vlan *rtnl_link_bridge_get_port_vlan(struct rtnl_link *l
...
+
+	entry = find_mst_state_entry(bd, instance);
+
+	if (entry == NULL)
+		return -NLE_OBJ_NOTFOUND;
+
+	mst_state_entry_del(entry);
+
+	if (nl_list_empty(&bd->mst_list))
+		bd->ce_mask &= ~BRIDGE_ATTR_MST;
```
thom311, line 1738: I don't think that the mask needs to be unset.

If you receive an object from kernel, then delete all entries and send it back, I think the fact that the `mst_list` is explicitly empty is important (and distinct from being unset).

Or not?

Same in `rtnl_link_bridge_clear_mst_port_state_info()`

Maybe calling delete should always explicitly set the flag on success (although here, we would expect that the flag was already set). This would be more a question for `rtnl_link_bridge_clear_mst_port_state_info()`.

lib/route/link/bridge.c
```diff
@@ -1376,6 +1625,186 @@ struct rtnl_link_bridge_vlan *rtnl_link_bridge_get_port_vlan(struct rtnl_link *l
...
+ * instance ID, state (BR_STATE_*), and an optional user provided data argument.
+ * MST entries should not be added or removed by the callback function.
+ *
+ * @return 0 on success or a negative error code
+ * @return -NLE_INVAL 		link or cb is NULL
+ * @return -NLE_OPNOTSUP 	Link is not a bridge
+ */
+int rtnl_link_bridge_foreach_mst_entry(
+	struct rtnl_link *link,
+	void (*cb)(uint16_t instance, uint8_t state, void *arg), void *arg)
```
thom311, line 1787: hm. This seems really cumbersome API... and also error prone, because the callback must guarantee not to modify the object while iterating... but ok. No better idea.

lib/route/link.c
```diff
@@ -1397,7 +1397,8 @@ int rtnl_link_build_get_request(int ifindex, const char *name,
 {
 	_nl_auto_nl_msg struct nl_msg *msg = NULL;
 	struct ifinfomsg ifi;
-	__u32 vf_mask = RTEXT_FILTER_VF;
+	__u32 vf_mask = RTEXT_FILTER_VF; // Shouldn't we also fill this in like
+					 // link_request_update does?
```
ronand-atl, line 1401: Actually it looks like sending an RTM_GETLINK message of type PF_UNSPEC would not return bridge specific attributes anyway. PF_BRIDGE seems to be needed for that. So it seems that getting MST entries is only possible when a cache is built first, specifying the family AF_BRIDGE.

lib/route/link/bridge.c
```diff
@@ -107,22 +117,110 @@ static inline struct bridge_data *bridge_data(struct rtnl_link *link)
...
+
+	nl_init_list_head(&dst_bd->mst_list);
+
+	nl_list_for_each_entry(entry, &src_bd->mst_list, list_node) {
+		struct mst_state_entry *entry_copy =
+			mst_state_entry_clone(entry);
+
+		if (!entry_copy) {
+			mst_list_clear(&dst_bd->mst_list);
+			free(dst_bd);
```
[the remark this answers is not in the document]
ronand-atl: Got it. I might have been hesitant because that required passing the link object as well (which I find odd anyway). The link that is passed is the partially cloned link which the bridge data will soon belong to. In practice the link parameter is ignored anyway but I'm unsure of the intended use for it.

lib/route/link/bridge.c
ronand-atl: I'm not entirely sure what you mean, could you explain some more? The nla_nest_cancel calls at the bottom will only cancel attributes that are not null, so it should work. I could change the code to the following which might be more clear:
```c
static int bridge_fill_mst(struct nl_msg *msg, struct nl_list_head *mst_list) {
	struct nlattr *attr = NULL;
	struct nlattr *entry_attr = NULL;
	struct mst_state_entry *entry;

	if (nl_list_empty(mst_list))
		return 0;

	attr = nla_nest_start(msg, IFLA_BRIDGE_MST);
	if (!attr)
		goto err_out;

	nl_list_for_each_entry(entry, mst_list, list_node) {
		entry_attr = nla_nest_start(msg, IFLA_BRIDGE_MST_ENTRY);
		if (!entry_attr)
			goto err_out_nest_cancel_attr;

		NLA_PUT_U16(msg, IFLA_BRIDGE_MST_ENTRY_MSTI, entry->msti);
		NLA_PUT_U8(msg, IFLA_BRIDGE_MST_ENTRY_STATE, entry->state);

		nla_nest_end(msg, entry_attr);
	}

	nla_nest_end(msg, attr);

	return 0;

nla_put_failure:
	nla_nest_cancel(msg, entry_attr);
err_out_nest_cancel_attr:
	nla_nest_cancel(msg, attr);
err_out:
	return -NLE_MSGSIZE;
}
```

ronand-atl: I went ahead and made this change since I think it's clearer.

lib/route/link/bridge.c
ronand-atl: > If you receive an object from kernel, then delete all entries and send it back, I think the fact that the mst_list is explicitly empty is important (and distinct from being unset).

No, the kernel will reject an empty list being sent so we don't send the list if it's empty. Also, the list can just contain a subset of MST instances to modify the state of for the port; the ones not listed will just not be changed. Actually setting BRIDGE_ATTR_MST doesn't serve much purpose since we can just check if the list is non-empty, but I added it for consistency (maybe I could remove it?).

lib/route/link/bridge.c
ronand-atl: This should be equivalent to using two different variables for each hash and comparing them at the end because of the way XOR works. This relies on each list not having duplicate items which is true.

```c
	nl_list_for_each_entry(entry, &a->mst_list, list_node) {
		hash_1 ^= _MST_HASH(entry->msti, entry->state);
	}
	nl_list_for_each_entry(entry, &b->mst_list, list_node) {
		hash_2 ^= _MST_HASH(entry->msti, entry->state);
	}
#undef _MST_HASH
	diff |= _DIFF(BRIDGE_ATTR_MST, hash_1 != hash_2);
```

I'm kind of hesitant about the hash method altogether, since there's a small chance of false positives or false negatives. I could use two nested for loops or keep the lists sorted so comparison is easier.

lib/route/link/bridge.c
ronand-atl: I could use a (thread local) static variable that acts as a lock that is set when this function is called and unset when the function is exited. The modification functions could then bail out if the variable is set. There is also `nl_list_for_each_entry_safe` which maybe I should be using instead.

lib/route/link/bridge.c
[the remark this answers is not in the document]
thom311: let's do
```diff
diff --git i/lib/route/link/bridge.c w/lib/route/link/bridge.c
index 94a29b1df6a7..0d7af5ba704a 100644
--- i/lib/route/link/bridge.c
+++ w/lib/route/link/bridge.c
@@ -170,12 +170,17 @@ static void mst_list_clear(struct nl_list_head *mst_list)
 		mst_state_entry_del(entry);
 }
 
+static void bridge_data_free(struct bridge_data *bd)
+{
+	mst_list_clear(&bd->mst_list);
+	free(bd);
+}
+
 static void bridge_free(struct rtnl_link *link, void *data)
 {
 	struct bridge_data *bd = data;
 
-	mst_list_clear(&bd->mst_list);
-	free(bd);
+	bridge_data_free(bd);
 }
 
 static struct mst_state_entry *find_mst_state_entry(struct bridge_data *bd,
@@ -216,7 +221,7 @@ static void *bridge_clone(struct rtnl_link *link, void *data)
 			mst_state_entry_clone(entry);
 
 		if (!entry_copy) {
-			bridge_free(link, dst_bd);
+			bridge_data_free(dst_bd);
 			return NULL;
 		}
 
```

lib/route/link/bridge.c
thom311: ah, I didn't realize that `nla_nest_cancel()` handles a `NULL` argument. So it was correct previously.

I still like this better.

lib/route/link/bridge.c
thom311: my point was that the equality check via the hash can have false negatives (wrongly not detecting inequality).

I don't think you need to do a order-independent comparison. While in some sense such configurations may be equivalent, if the order differs, we can assume that the overall object is different. That simplifies the implementation, and I probably would expect to consider ordering in the comparison.

lib/route/link/bridge.c
thom311: ok, it's good then.

While BRIDGE_ATTR_MST is redundant (to just checking whether the list is empty), at various places we have also such flags when we could deduce via alternative ways whether the value is unset. I think it's fine to have it.

lib/route/link/bridge.c
thom311: hm...

I general, I find the way we track the mst_state_entry via a linked list not great. It has more overhead than just allocating (and `realloc()`) an array. Since we cannot do efficient O(1) lookup by index, it also does not lean to an API where we could have
```C
int rtnl_link_bridge_get_mst_entry_at(struct rtnl_link *link, size_t index, uint16_t *out_msti, uint8_t *out_state);
```

I think I am fine with it as-is.

ronand-atl: There is no significance to the order of the MST entries so I don't think there is much use in exposing a way to get them by index. I chose to just use a list because that's what some other modules use, and it made implementation easier. However I do agree that other data structures could be better suited (such as a hashmap or self-balancing tree). If the list implementation is outgrown, the implementation could be switched to one of these without changing the external API, since the list implementation is not exposed to the user.

thom311: > There is no significance to the order of the MST entries so I don't think there is much use in exposing a way to get them by index.

While the index has no particular meaning in most regards, it would have a meaning for accessing the elements.

Just like the callback is invoked in a certain order. That order is also not particularly meaningful, but such order still exists and is visible to the user of the API.

Anyway.

thom311:
looks good to me. Merged as b75e27deb8957a04fd84539cd9a1f772930d17d2

Thank you for the patch!

PR #412, opened 2024-11-14
[th/c-list] import c-list library as third-party module for internal use

thom311:
We already have an internal implementation for a circular, intrusive,
doubly-linked list (in `include/netlink/list.h`).

However, I think the "c-list" library is excellent, and in 2024 a
netlink library should no longer reimplement a basic data structure.

Vendor "c-list" via git-subtree. We want to strictly follow upstream (no
local changes on our side) but re-import the library when there are new
upstream changes.

Existing users of <netlink/list.h> are not required to be rewritten. But
new users should use "c-list.h". This is only an internal dependency, we
anyway wouldn't want to expose such a list in our public API.

[1] https://github.com/c-util/c-list

Imported via:

    $ git subtree add --prefix third_party/c-list/ git@github.com:c-util/c-list.git main --squash

Update with:

    $ git subtree pull --prefix third_party/c-list/ git@github.com:c-util/c-list.git main --squash

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -338,6 +338,9 @@ noinst_HEADERS = \
 	include/nl-priv-dynamic-core/object-api.h \
 	include/nl-priv-dynamic-route/nl-priv-dynamic-route.h \
 	include/nl-priv-static-route/nl-priv-static-route.h \
+	\
+	third_party/c-list/src/c-list.h \
+	\
 	$(NULL)
 
 ###############################################################################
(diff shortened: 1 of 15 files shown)


ronand-atl:
LGTM :+1: 

Could it be good to mention in the existing list implementation that new modules should prefer to use this implementation instead?

thom311:
> Could it be good to mention in the existing list implementation that new modules should prefer to use this implementation instead?

This is mentioned both in the commit message and in the PR description here. Do you have anything else in mind?

ronand-atl:
Probably just a note at the top of the old `list.h` file. Sorry I should have specified that.

PR #413, opened 2024-11-24
xfrm: Add support for xfrm interface ID in xfrm state

carlgsmith:


libnl-xfrm-3.sym
```diff
@@ -82,6 +82,7 @@ global:
 	xfrmnl_sa_get_encap_tmpl;
 	xfrmnl_sa_get_family;
 	xfrmnl_sa_get_flags;
+	xfrmnl_sa_get_if_id;
```
thom311, line 85: new entries must move to the bottom of the file, to a new `libnl3_12_0` section. See the comment on top of the file.

lib/xfrm/sa.c
```diff
@@ -1700,6 +1713,22 @@ int xfrmnl_sa_set_family (struct xfrmnl_sa* sa, unsigned int family)
 	return 0;
 }
 
+int xfrmnl_sa_get_if_id (struct xfrmnl_sa* sa)
+{
+	if (sa->ce_mask & XFRM_SA_ATTR_IF_ID)
+		return sa->if_id;
+	else
+		return -1;
```
thom311, line 1721: return `-NLE_NOATTR`.

also, return-early and check for errors first:
```C
if (!(sa->ce_mask & XFRM_SA_ATTR_IF_ID)
     return -NLE_NOATTR;
return sa->if_id;
```

I understand that you copy&pasted this code, but we shouldn't follow bad patterns.

lib/xfrm/sa.c
```diff
@@ -1006,6 +1010,11 @@ int xfrmnl_sa_parse(struct nlmsghdr *n, struct xfrmnl_sa **result)
 		sa->ce_mask |= XFRM_SA_ATTR_OFFLOAD_DEV;
 	}
 
+	if (tb[XFRMA_IF_ID]) {
+		sa->if_id      =   *(uint32_t*)nla_data(tb[XFRMA_IF_ID]);
+		sa->ce_mask |= XFRM_SA_ATTR_IF_ID;
```
thom311, line 1015: you must add an entry to `xfrm_sa_policy` to ensure that the provided data is large enough to hold a uint32.

lib/xfrm/sa.c
```diff
@@ -1006,6 +1010,11 @@ int xfrmnl_sa_parse(struct nlmsghdr *n, struct xfrmnl_sa **result)
 		sa->ce_mask |= XFRM_SA_ATTR_OFFLOAD_DEV;
 	}
 
+	if (tb[XFRMA_IF_ID]) {
+		sa->if_id      =   *(uint32_t*)nla_data(tb[XFRMA_IF_ID]);
```
thom311, line 1014: let's use `nla_get_u32()`.

thom311 (changes requested):

libnl-xfrm-3.sym
carlgsmith: Done. libnl_3_6 to libnl_3_12

lib/xfrm/sa.c
carlgsmith: Done. Also change to return value via passed in pointer to ensure we get full uint32_t

lib/xfrm/sa.c
carlgsmith: Done.

lib/xfrm/sa.c
carlgsmith: Done.

carlgsmith:
Also added the policy versions of if_id as they are required as well.

lib/xfrm/sa.c
thom311: oh right, this was important. Please change the `unsigned int` parameters in the public API to `uint32_t` types.

carlgsmith: Done. All public API use of if_id is now uint32_t

thom311 (approved):

thom311:
nice! Thank you.

merged as 49518ca56d8b632fca4dd17834e62e7d7ee85fa1

Issue #377, opened 2024-04-24
bridge object in link cache may get replaced by an incomplete object


KanjiMonster:
I observed the link cache sometimes returning an incomplete object of family `AF_BRIDGE`, but with no link type set (and thus any rtnl_link_bridge_* call triggering an assert).

This happens shortly after creation and attaching a port to a bridge.

Running with `NLCB=debug`, the following messages can be seen:

complete bridge object ("good" kernel message) => link cache returns a full object

```
-- Debug: Received Message: 
--------------------------   BEGIN NETLINK MESSAGE ---------------------------
  [NETLINK HEADER] 16 octets
    .nlmsg_len = 1864 
    .type = 16 <route/link::new>
    .flags = 0 <>
    .seq = 0 
    .port = 0
  [PAYLOAD] 16 octets 
    00 00 01 00 40 00 00 00 03 10 00 00 00 00 00 00 ....@...........
  [ATTR 03] 9 octets
    73 77 62 72 69 64 67 65 00                      swbridge.
  [PADDING] 3 octets
    00 00 00                                        ...
  [ATTR 13] 4 octets
    e8 03 00 00                                     ....
  [ATTR 16] 1 octets
    02                                              .
  [PADDING] 3 octets
    00 00 00                                        ...
  [ATTR 17] 1 octets
    00                                              .
  [PADDING] 3 octets
    00 00 00                                        ...
  [ATTR 04] 4 octets
    dc 05 00 00                                     ....
  [ATTR 50] 4 octets
    44 00 00 00                                     D...
  [ATTR 51] 4 octets
    ff ff 00 00                                     ....
  [ATTR 27] 4 octets
    00 00 00 00                                     ....
  [ATTR 30] 4 octets
    00 00 00 00                                     ....
  [ATTR 61] 4 octets
    00 00 00 00                                     ....
  [ATTR 31] 4 octets
    01 00 00 00                                     ....
  [ATTR 40] 4 octets
    ff ff 00 00                                     ....
  [ATTR 41] 4 octets
    00 00 01 00                                     ....
  [ATTR 58] 4 octets
    00 00 01 00                                     ....
  [ATTR 63] 4 octets
    00 00 01 00                                     ....
  [ATTR 64] 4 octets
    00 00 01 00                                     ....
  [ATTR 59] 4 octets
    00 00 01 00                                     ....
  [ATTR 60] 4 octets
    ff ff 00 00                                     ....
  [ATTR 32] 4 octets
[...]

PR #414, opened 2024-11-26
route/link: support update operation for link objects

viktor-iarmola:
When bridge interfaces are in vlan-aware mode, one of the rtnetlink updates they produce was not handled correctly.

When the set of vlans active on a vlan-aware bridge itself it being changed it produces RTM_NEWLINK message with AF_BRIDGE family (e.g. with `bridge vlan add dev br0 vid 3 self`).
Unlike other messages arriving with AF_BRIDGE family - this one doesn't refer to one of the member interfaces, but to the bridge itself as if it was its own member.

This message doesn't contain full information about bridge - mostly about vlans in IFLA_AF_SPEC attribute. It does, however, refer to the bridge interface itself via ifi_index parameter.

`libnl` parses such message into a link object with `(AF_BRIDGE, br-ifindex)` key.

At the same time, the regular rtnetlink message describing bridge interface itself (one that arrives with AF_UNSPEC family and IFLA_LINKINFO/IFLA_INFO_KIND == bridge) will also be parsed into link object with the same key `(AF_BRIDGE, br-ifindex)`. That is caused by the libnl classifying it internally as bridge (via IFLA_INFO_KIND) and overriding its family to AF_BRIDGE.

Such collision results in an unexpected behavior when we have a cache with the information about bridge (from AF_UNSPEC message) and then receive notification produced by `bridge vlan add/del ... self`. As both link entries have the same key - new one will override the old - and we will have a misleading entry in cache, which doesn't contain most of the information we've had about the bridge. It also defies some established expectations about bridge interfaces in libnl. E.g. `rtnl_link_is_bridge` is 1 for such entries, but `rtnl_link_get_type` is NULL.

This situation can be amended by implementing update mechanism for link objects. With that, it is possible to properly handle such special rtnetlink messages and merge information available in them into full bridge description.

Addresses #377

### DIFF
--- a/lib/route/link.c
+++ b/lib/route/link.c
@@ -1126,6 +1126,20 @@ static void link_keygen(struct nl_object *obj, uint32_t *hashkey,
 	return;
 }
 
+static int link_update(struct nl_object *old_obj, struct nl_object *new_obj)
+{
+	struct rtnl_link *src = (struct rtnl_link *) new_obj;
+	struct rtnl_link *dst = (struct rtnl_link *) old_obj;
+
+	if (!dst->l_af_ops)
+		return -NLE_OPNOTSUPP;
+
+	if (dst->l_af_ops->ao_update)
+		return dst->l_af_ops->ao_update(dst, src);
+
+	return -NLE_OPNOTSUPP;
+}
+
 static uint64_t link_compare(struct nl_object *_a, struct nl_object *_b,
 			     uint64_t attrs, int flags)
 {
@@ -3215,6 +3229,7 @@ static struct nl_object_ops link_obj_ops = {
 	},
 	.oo_compare		= link_compare,
 	.oo_keygen		= link_keygen,
+	.oo_update		= link_update,
 	.oo_attrs2str		= link_attrs2str,
 	.oo_id_attrs		= LINK_ATTR_IFINDEX | LINK_ATTR_FAMILY,
 };
--- a/lib/route/link/bridge.c
+++ b/lib/route/link/bridge.c
@@ -973,6 +973,29 @@ static int bridge_compare(struct rtnl_link *_a, struct rtnl_link *_b,
 }
 /** @endcond */
 
+static int bridge_update(struct rtnl_link *old_obj, struct rtnl_link *new_obj)
+{
+	struct bridge_data *dst = bridge_data(old_obj);
+	struct bridge_data *src = bridge_data(new_obj);
+
+	/*
+	 * Update only when we're dealing with a case of bridge
+	 * notification on changes to itself as a bridge port.
+	 * E.g. after `bridge vlan add dev br0 vid 3 self`
+	 * Such notifications don't have full bridge details
+	 * and thus have to be merged with parent rather than replace it.
+	 * Reliable criterion to detect the case we're interested in
+	 * is to look if there is IFLA_MASTER attr equal to ifi_index
+	 * in the same message.
+	 */
+	if (new_obj->l_index != new_obj->l_master)
+		return -NLE_OPNOTSUPP;
+
+	*dst = *src;
+
+	return NLE_SUCCESS;
+}
+
 /**
  * Allocate link object of type bridge
  *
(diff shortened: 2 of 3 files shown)


thom311:
It seems hard to understand all the consequence of this patch. But the reasoning and the patch looks good to me, so unless somebody objects, I'll take it... (let's wait a bit).

lib/route/link.c
```diff
@@ -1126,6 +1126,20 @@ static void link_keygen(struct nl_object *obj, uint32_t *hashkey,
...
+	struct rtnl_link *src = (struct rtnl_link *) new_obj;
+	struct rtnl_link *dst = (struct rtnl_link *) old_obj;
+
+	if (!dst->l_af_ops)
+		return -NLE_OPNOTSUPP;
+
+	if (dst->l_af_ops->ao_update)
+		return dst->l_af_ops->ao_update(dst, src);
+
+	return -NLE_OPNOTSUPP;
```
thom311, line 1140: let's return early, so we handle failure (OPNOTSUPP) cases first. Also, add a code comment like:

```C
        struct rtnl_link *src = (struct rtnl_link *) new_obj;
        struct rtnl_link *dst = (struct rtnl_link *) old_obj;
        int r;

        if (!dst->l_af_ops)
		return -NLE_OPNOTSUPP;

	if (!dst->l_af_ops->ao_update)
		return -NLE_OPNOTSUPP;

        r = dst->l_af_ops->ao_update(dst, src);
        if (r < 0)
            return r;

        /* dst was already modified. We can no longer fail at this point. */

	return 0;
```

KanjiMonster:
I'm not sure if you can just replace the bridge data, as the bridge data may be quite incomplete, see https://github.com/thom311/libnl/issues/377 for an example.

viktor-iarmola:
>I'm not sure if you can just replace the bridge data, as the bridge data may be quite incomplete,

Replacing bridge data entirely in the link entry for bridge interfaces doesn't look wrong to me - primarily because bridge interfaces cannot be members of other bridges. As such, only a single operation (`bridge vlan add/del dev brX vid Y self`) can touch that info as of now - and there are no other sources for it.

That said, the approach is indeed not general enough and updated bridge data will be lost whenever a new AF_UNSPEC update for bridge itself comes (https://github.com/thom311/libnl/pull/414#issuecomment-2503775634). 
I do hope to make an improved version of it in the future - but, practically speaking, this one is just fine for the issue at hand.

KanjiMonster:
> > I'm not sure if you can just replace the bridge data, as the bridge data may be quite incomplete,
> 
> Replacing bridge data entirely in the link entry for bridge interfaces doesn't look wrong to me - primarily because bridge interfaces cannot be members of other bridges. As such, only a single operation (`bridge vlan add/del dev brX vid Y self`) can touch that info as of now - and there are no other sources for it.

Not sure what the bridges (not) being members of other bridges have to do with it.

What I mean is that the bridge_data in the AF_BRIDGE update from #377 *only* contains the vlan info, but not e.g. the MST info - which now would get lost on replacement. Which would also be a memory leak, since this is a linked list of allocated objects.

So you would need to update/replace the old bridge_data's vlan info etc with the values from the AF_BRIDGE update, but not touch anything that isn't part of the minimal update.

viktor-iarmola:
@KanjiMonster 

>bridge_data in the AF_BRIDGE update from https://github.com/thom311/libnl/issues/377 only contains the vlan info, but not e.g. the MST info - which now would get lost on replacement.

I see. You're right, some fields in `struct bridge_data` can't just be overwritten. I overlooked that because, to the best of my knowledge, mst information is not sent by kernel for AF_BRIDGE updates about bridge itself - so there can't be a leak.
That said, this assumptions could be wrong or kernel can decide to start sending it.

Thank you for pointing this out, it needs to be addressed.

viktor-iarmola:
>Not sure what the bridges (not) being members of other bridges have to do with it.

Most of the contents of regular AF_BRIDGE updates have to do with the interface status as "port in some bridge". Since bridges can't be members of other bridges - most of the contents of `struct bridge_data` will never be filled for bridges themselves. Vlan info is kind of outlier here and probably the only valid member of `struct bridge_data` for bridges.

KanjiMonster:
> > Not sure what the bridges (not) being members of other bridges have to do with it.
> 
> Most of the contents of regular AF_BRIDGE updates have to do with the interface status as "port in some bridge". Since bridges can't be members of other bridges - most of the contents of `struct bridge_data` will never be filled for bridges themselves. Vlan info is kind of outlier here and probably the only valid member of `struct bridge_data` for bridges.

Be aware that any AF_UNSPEC updates for links with the IFLA_LINK_TYPE "bridge" are "fixed up" to have the family AF_BRIDGE.

This is happening in https://github.com/thom311/libnl/blob/main/lib/route/link.c#L683-L698.

PR #416, opened 2024-12-17
cache: cache_include(): fix double put for cloned objects

KanjiMonster:
When switching to auto obj, a nl_object_put() was left inplace in cache_include(). This leads to a double reference drop of the cloned object, leading to use after free later or triggering an assert that the reference count went negative.

Fixes: 831e98688a2f ("cache: use the new _nl_auto_nl_object helper")

### DIFF
--- a/lib/cache.c
+++ b/lib/cache.c
@@ -807,7 +807,6 @@ static int cache_include(struct nl_cache *cache, struct nl_object *obj,
 				if (cb_v2) {
 					cb_v2(cache, clone, old, diff,
 					      NL_ACT_CHANGE, data);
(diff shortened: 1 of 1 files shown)


thom311:
thanks for the fix!!

hm, this is pretty bad. I think I should do a 3.12 release soon, just for this.

thom311:
I understood that it's only relevant with the v2 callback.

But I don't have a clear picture who uses that. It was unnoticed since:

- introduces in August 2024 (arguably, few people test upstream before it gets released)
- 3.11 release, ~47 days ago

ok, then maybe no immediate hurry... :)

PR #420, opened 2025-02-05
cache: add nl_cache_resync_v2()

KanjiMonster:
When the include callback v2 support was added in 66d032ad443a ("cache_mngr: add include callback v2"), resync_cb() was updated to handle both old and v2 callbacks, but no actual nl_cache_resync_v2() was added to make use of it.

Fix this by adding an appropriate implementation.


thom311:
lgtm. Merged.

Thank you!!

PR #421, opened 2025-02-18
Change vlan module to set QOS mapping flag

EriComic:
The newer versions of the Linux kernel now check to ensure that QOS mappings also have IFLA_VLAN_QOS_MAPPING set [1]. The kernel will discard QOS atributes where this is not set.

The current libnl code for setting vlan attributes uses variable i to set this parameter. This means that only the second QOS map will be applied since i = 1 = IFLA_VLAN_QOS_MAPPING, the rest will be discarded.

This patch replaces i in the PUT macro to use the already existing enum value IFLA_VLAN_QOS_MAPPING which equals 1.

[1] https://github.com/torvalds/linux/commit/6c21660fe221a15c789dee2bc2fd95516bc5aeaf 

Co-authored-by: Daniel Falk falk@haleytek.com

### DIFF
--- a/lib/route/link/vlan.c
+++ b/lib/route/link/vlan.c
@@ -328,7 +328,7 @@ static int vlan_put_attrs(struct nl_msg *msg, struct rtnl_link *link)
 				map.from = i;
 				map.to = vi->vi_ingress_qos[i];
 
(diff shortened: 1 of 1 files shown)


thom311:
lgtm. Merged.

Thank you!

PR #426, opened 2025-06-03
bridge_info: implement missing rtnl_link_info_ops::io_clone()

KanjiMonster:
Link info ops that use link info also need to implement `rtnl_link_info_ops::io_clone()`, else cloning the link will result in a broken object, and trying to use any `rtnl_link_bridge_*()` accessors will result in a null pointer access due to `rtnl_link::l_info` not being taken over.

So implement the missing `rtnl_link_info_ops::io_clone()` callback.

Fixes: 7391a38e3d5e ("bridge: Add support for link_info of a bridge")

### DIFF
--- a/lib/route/link/bridge_info.c
+++ b/lib/route/link/bridge_info.c
@@ -88,6 +88,26 @@ static int bridge_info_alloc(struct rtnl_link *link)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


thom311:
merged. Thank you!!

PR #427, opened 2025-06-03
bridge_info: implement missing rtnl_link_info_ops::io_compare()

KanjiMonster:
Without rtnl_link_info_ops::io_compare() changes in link info are not picked up by the cache and no notificatons are sent out as rtnl_link_info_data_compare() returns 0 when the callback is missing.

Fix this by implementing the missing io_compare() for bridge_info.

This allows e.g. changes in ageing time to be seen.

Fixes: 7391a38e3d5e ("bridge: Add support for link_info of a bridge")

### DIFF
--- a/lib/route/link/bridge_info.c
+++ b/lib/route/link/bridge_info.c
@@ -261,6 +261,53 @@ static void bridge_info_free(struct rtnl_link *link)
 	_nl_clear_free(&link->l_info);
 }
 
+static int bridge_info_compare(struct rtnl_link *link_a,
+			       struct rtnl_link *link_b, int flags)
+{
+	struct bridge_info *a = link_a->l_info;
+	struct bridge_info *b = link_b->l_info;
+	uint32_t attrs = flags & LOOSE_COMPARISON ? b->ce_mask : ~0u;
+	int diff = 0;
+
+#define _DIFF(ATTR, EXPR) ATTR_DIFF(attrs, ATTR, a, b, EXPR)
+	diff |= _DIFF(BRIDGE_ATTR_VLAN_FILTERING,
+		      a->b_vlan_filtering != b->b_vlan_filtering);
+	diff |= _DIFF(BRIDGE_ATTR_VLAN_PROTOCOL,
+		      a->b_vlan_protocol != b->b_vlan_protocol);
+	diff |= _DIFF(BRIDGE_ATTR_VLAN_STATS_ENABLED,
+		      a->b_vlan_stats_enabled != b->b_vlan_stats_enabled);
+	diff |= _DIFF(BRIDGE_ATTR_AGEING_TIME,
+		      a->b_ageing_time != b->b_ageing_time);
+	diff |= _DIFF(BRIDGE_ATTR_VLAN_DEFAULT_PVID,
+		      a->b_vlan_default_pvid != b->b_vlan_default_pvid);
+	diff |= _DIFF(BRIDGE_ATTR_NF_CALL_IPTABLES,
+		      a->b_nf_call_iptables != b->b_nf_call_iptables);
+	diff |= _DIFF(BRIDGE_ATTR_NF_CALL_IP6TABLES,
+		      a->b_nf_call_ip6tables != b->b_nf_call_ip6tables);
+	diff |= _DIFF(BRIDGE_ATTR_NF_CALL_ARPTABLES,
+		      a->b_nf_call_arptables != b->b_nf_call_arptables);
+	diff |= _DIFF(BRIDGE_ATTR_STP_STATE, a->b_stp_state != b->b_stp_state);
+	diff |= _DIFF(BRIDGE_ATTR_MCAST_ROUTER,
+		      a->b_mcast_router != b->b_mcast_router);
+	diff |= _DIFF(BRIDGE_ATTR_MCAST_SNOOPING,
+		      a->b_mcast_snooping != b->b_mcast_snooping);
+
+	if (flags & LOOSE_COMPARISON)
+		diff |= _DIFF(
+			BRIDGE_ATTR_BOOLOPT,
+			(b->b_boolopts.optmask & ~a->b_boolopts.optmask) ||
+				((a->b_boolopts.optval ^ b->b_boolopts.optval) &
+				 b->b_boolopts.optmask));
+	else
+		diff |= _DIFF(
+			BRIDGE_ATTR_BOOLOPT,
+			(a->b_boolopts.optmask != b->b_boolopts.optmask) ||
+				(a->b_boolopts.optval != b->b_boolopts.optval));
+#undef _DIFF
+
+	return diff;
+}
+
 static struct rtnl_link_info_ops bridge_info_ops = {
 	.io_name = "bridge",
 	.io_alloc = bridge_info_alloc,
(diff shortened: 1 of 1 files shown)


KanjiMonster:
I'm not 100% confident about the `b_boolopts` comparison. It has a valmask and valopts attributes, but for now I opted to only compare the valopts. I assume the kernel will always notify about all of them, so the valmask will be essentially static during runtime.

KanjiMonster:
Updated to follow the way how https://github.com/thom311/libnl/blob/main/lib/route/addr.c#L565-L569 handles the `LOOSE_COMPARISON`.

lib/route/link/bridge_info.c
```diff
@@ -261,13 +261,56 @@ static void bridge_info_free(struct rtnl_link *link)
...
+		      a->b_mcast_router != b->b_mcast_router);
+	diff |= _DIFF(BRIDGE_ATTR_MCAST_SNOOPING,
+		      a->b_mcast_snooping != b->b_mcast_snooping);
+	if (flags & LOOSE_COMPARISON)
+		diff |= _DIFF(BRIDGE_ATTR_BOOLOPT,
+			      (a->b_boolopts.optval ^ b->b_boolopts.optval) &
+				      b->b_boolopts.optmask);
+	else
+		diff |= _DIFF(BRIDGE_ATTR_BOOLOPT,
+			      a->b_boolopts.optval != b->b_boolopts.optval);
```
thom311, line 300: I think it also should compare `b_boolopts.optmask`.

lib/route/link/bridge_info.c
```diff
@@ -261,13 +261,56 @@ static void bridge_info_free(struct rtnl_link *link)
...
+		      a->b_nf_call_arptables != b->b_nf_call_arptables);
+	diff |= _DIFF(BRIDGE_ATTR_STP_STATE, a->b_stp_state != b->b_stp_state);
+	diff |= _DIFF(BRIDGE_ATTR_MCAST_ROUTER,
+		      a->b_mcast_router != b->b_mcast_router);
+	diff |= _DIFF(BRIDGE_ATTR_MCAST_SNOOPING,
+		      a->b_mcast_snooping != b->b_mcast_snooping);
+	if (flags & LOOSE_COMPARISON)
+		diff |= _DIFF(BRIDGE_ATTR_BOOLOPT,
+			      (a->b_boolopts.optval ^ b->b_boolopts.optval) &
+				      b->b_boolopts.optmask);
```
thom311, line 297: `& a->b_boolopts.optmask` ?

thom311: If we ignore `a->b_boolopts.optmask`, then the code honors bits of `a->b_boolopts.optval` that are not explicitly set to a meaningful value. I think such bits should not be honored, which can be achieved via `& a->b_boolopts.optval`.

If we want something asymmetric where that `b` is a filter object, then when we search for certain `b->b_boolopts.optmask` bits, then missing bits in `a` must result in a difference. Something like:
```
    ((a->b_boolopts.optval ^ b->b_boolopts.optval) & a->b_boolopts.optmask & b->b_boolopts.optmask) 
    || (b->b_boolopts.optmask & ~a->b_boolopts.optmask)
```

If other types don't do that, it seems a bug to me.

WDYT?

KanjiMonster: I think it makes sense to only look a b's `ce_mask`/`optmask` in the case of `LOOSE_COMPARISON`.

Let's say you want all bridges with a certain flag set, so you set `optmask` to include that bit and `optval` accordingly. If there are bridges that do not have that bit set in their optmask, then the filter would also match those objects, which is IMHO wrong (since we wouldn't compare it anymore for them).

Same for the other users, where b's `ce_mask` defines which values to compare. Say you wanted to find all links that have a certain device as the master. Then you call `rtnl_link_set_master(link, 123);` on your filter. All links without a master will miss the `LINK_ATTR_MASTER` bit in their `ce_mask`, but if we now skip checking the value because `a->ce_mask` does not have it, we will now also match all links without a master.

thom311: > I think it makes sense to only look a b's ce_mask/optmask in the case of LOOSE_COMPARISON.
> ...
> All links without a master will miss the LINK_ATTR_MASTER bit in their ce_mask, but if we now skip checking the value because a->ce_mask does not have it, we will now also match all links without a master.

You talk here about `ce_mask`, which is handled separately by `ATTR_DIFF()`. But as you say, if `b` asks to check for a `MASTER` attribute  that is not in `a->ce_mask`, then an `a` without a master must not match. That requires, that `ATTR_DIFF()` looks at `a->ce_mask` in some way.

I am saying, that `ce_mask` and `b_boolopt.optmask` must behave similar. You must check for a missing `a->b_boolopt.optmask` to get what you just said. The current code
```
+                 diff |= _DIFF(BRIDGE_ATTR_BOOLOPT,
+                                (a->b_boolopts.optval ^ b->b_boolopts.optval) &
+                                         b->b_boolopts.optmask);
```
completely ignores `a->b_boolopts.optmask`, which is wrong.

I suggested two options what to do about missing `a->b_boolopts.optmask` . Either accept or reject the `a` object. The benefit of ignoring the missing options is that it makes the comparison symmetric. But if we assume that `b` is a filter object, then it makes sense to do something asymmetric. In any case,  `a->b_boolopt.optmask` must be considered.


These bits have three states: undefined (unset the optmask), unset (explicilty set to zero) and set.

If you treat `b` as the lookup/filter object and we do an asymatric match, then it should look like:

```
        | a undef  | a unset | a set
----------------------------------------
b undef | match    | match   | match
b unset | nomatch  | match   | nomatch
b set   | nomatch  | nomatch | match
```

I think, you get this with
```
    ((a->b_boolopts.optval ^ b->b_boolopts.optval) & a->b_boolopts.optmask & b->b_boolopts.optmask) 
    || (b->b_boolopts.optmask & ~a->b_boolopts.optmask)
```

thom311: With
```C
                 diff |= _DIFF(BRIDGE_ATTR_BOOLOPT,
                                (a->b_boolopts.optval ^ b->b_boolopts.optval) &
                                         b->b_boolopts.optmask);
```
you implement
```
        | a undef  | a unset | a set
----------------------------------------
b undef | match    | match   | match
b unset | match*   | match   | nomatch
b set   | nomatch* | nomatch | match
```

The `match*`/`nomatch*` is under the assumption that `a->b_boolopts.optval` has the bit unset, if the `b_boolopts.optmask` is unset. That is, kernel sets undefined flags to zero. Which is an additional assumption, that would be nice to avoid.

thom311: Forgive me, I would prefer to not bikeshed about this, but...

> Especially the non-loose comparison is now dependent on the kernel version if trying to compare kernel objects with manually created ones, since the kernel statically initializes optmask to all supported options.

The non-loose comparison now checks that all parts match. I think that is what we would expect from a is-equal method. If you want to lookup by a needle, then there is already the loose-comparison for that reason. 

> I still think it's fine to rely on the kernel initializing unknown values to zero

We don't need to trust kernel about doing that. We can just mask out the bits that are unset in the mask, and interpret the value the way you would expect. The result will be the same if kernel does what we expect. And it will be defined if kernel does not.

---

I suggested
```C
   ((a->b_boolopts.optval ^ b->b_boolopts.optval) & a->b_boolopts.optmask & b->b_boolopts.optmask) 
    || (b->b_boolopts.optmask & ~a->b_boolopts.optmask)
```
but you did
```C
                          (a->b_boolopts.optmask & ~b->b_boolopts.optmask) ||
                                   ((a->b_boolopts.optval ^ b->b_boolopts.optval) &
                                    b->b_boolopts.optmask)
```

Maybe I made a mistake in what I suggested earlier, so what you provided here is actually correct. Could you comment why you implemented something different, and why that is preferable? Thanks. I still think my earlier suggestion should be done. Or at least, please explain why not and why it differs.

thom311:
merged as 320b2bd

thank you!!

PR #432, opened 2025-07-21
ip6_tnl: Add API to mark tunnels to "collect metadata"

cpaasch-oai:
To set the flag IFLA_IPTUN_COLLECT_METADATA on the tunnel.

### DIFF
--- a/libnl-route-3.sym
+++ b/libnl-route-3.sym
@@ -1367,6 +1367,8 @@ global:
 
 libnl_3_12 {
 global:
+	rtnl_link_ip6_tnl_get_collect_metadata;
+	rtnl_link_ip6_tnl_set_collect_metadata;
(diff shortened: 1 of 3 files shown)


libnl-route-3.sym
```diff
@@ -1369,4 +1369,6 @@ libnl_3_12 {
 global:
 	rtnl_link_is_bond;
 	rtnl_nh_get_oif;
+	rtnl_link_ip6_tnl_get_collect_metadata;
+	rtnl_link_ip6_tnl_set_collect_metadata;
```
thom311, line 1373: could you please sort these lines to be first? Rest lgtm.

thom311:
merged as 0a611be8023af56a95c19be573cbe7c9378adaa3

thanks for the patch!!

PR #433, opened 2025-07-22
Add EHOSTUNREACH to nl_syserr2nlerr

cpaasch-oai:
For improved error-output.

### DIFF
--- a/include/netlink/errno.h
+++ b/include/netlink/errno.h
@@ -45,8 +45,9 @@ extern "C" {
 #define NLE_IMMUTABLE		32
 #define NLE_DUMP_INTR		33
 #define NLE_ATTRSIZE		34
+#define NLE_HOSTUNREACH		35
 
-#define NLE_MAX		NLE_ATTRSIZE
+#define NLE_MAX		NLE_HOSTUNREACH
 
 extern const char *	nl_geterror(int);
 extern void		nl_perror(int, const char *);
--- a/lib/error.c
+++ b/lib/error.c
@@ -43,6 +43,7 @@ static const char *errmsg[NLE_MAX+1] = {
 [NLE_IMMUTABLE]		= "Immutable attribute",
 [NLE_DUMP_INTR]		= "Dump inconsistency detected, interrupted",
 [NLE_ATTRSIZE]		= "Attribute max length exceeded",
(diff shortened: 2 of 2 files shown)


thom311:
hm, changing return/error codes is an API change/break. This is an unfortunate thing. Also unfortunate, to only notice the break after doing the release.

But, if we generally disallow such changes, it means the functions can never change to more detailed error codes. For example, also the change to introduce `NLE_NETDOWN` (fe74608d550e15cf407b30a7273d60d28e4fb6fd) is a no-go. That seems very limiting to freeze the API that way, and basically prevents "improvements".

The real issue is that we try to duplicate errno codes, but only do that for a selected few. What should have happend, is to map the entire range of errno 1..4096 to some `NLE_REAL_ERRNO_BASE+1 .. NLE_REAL_ERRNO_BASE+4096`. Or, alternatively, like systemd does, just stick to the original errno codes without NL specific extensions. But that is too late too.

What do you think? Should we revert those two patches (and do a 3.13 release in a few weeks)?

@KanjiMonster which function did you call, where the change in behavior causes problems?

> Now we need to special case if we build against 3.12 because we can't unconditionally check for the new error message

You can safely do
```
#define MY_NLE_HOSTUNREACH            35
#define MY_NLE_NETDOWN                36

...

r = nl_foo(...);
if (r == -NLE_FAILURE or r == -MY_NLE_HOSTUNREACH) ...
```
and don't need to check the version. That is regardless of whether we revert 45449e0bc6932c65be19674f4bc9db9ec99fa369, we will take care to not reuse error code `35` for anything else.

thom311:
I did now 3cc413f0a0b4e8de0bc056bf78068c1041dbb43c, which basically disables the effect of these two changes (#433, #441).

I will see to do a new release in a few weeks (waiting a bit longer to get possibly other bug reports first).

Unfortunately, that doesn't really solve the problem for the affected users. I think if you are affected, you should apply a workaround (that is, expect either `NLE_FAILURE` **OR** error code`35`/`36`). Or maybe you can wait it out and never use version 3.12.

I think what should be done is that certain functions guarantee to preserve the underlying `errno` when they fail (or only when they fail with `NLE_FAILURE`?). That way, you can do `if (err == -NLE_FAILURE && errno == EHOSTUNREACH) ...`. Not great, but maybe the least bad solution. TODO.

PR #434, opened 2025-07-24
Add support for resizeable hashtables and adopt them for the nl_cache

cpaasch-oai:
Currently, hashtable.c implements a static-sized hashtable (1024 by default). This works well if we know upfront how many elements we expect to support. However, a resizeable approach would be much better because it will dynamically resize the hashtable based on the load.

Adoption of resizeable hashtables significantly improves the time it takes to create & populate the cache. For example, for 500k ip nexthops, `nl-nh-list` call goes from ~12s to ~3s.

Because hashtable.h is a public API, we can't change this as is and rather add another variant of hashtable (not exposed as a public API).


include/nl-priv-dynamic-core/hashtable-api.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+#ifndef NETLINK_HASHTABLE_API_H_
+#define NETLINK_HASHTABLE_API_H_
+
+#include "nl-priv-dynamic-core/object-api.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
thom311, line 18: There is no need to have `extern "C"` here. 

This header is internal-only (only used to build libnl3 itself). Users must build libnl3 with a C compiler. And since other such headers also don't have `extern "C"`, you anyway couldn't use a C++ compiler to build libnl3.

thom311:
(more review will follow later)

include/nl-priv-dynamic-core/hashtable-api.h
```diff
@@ -0,0 +1,24 @@
+/* SPDX-License-Identifier: LGPL-2.1-only */
+/*
+ * Private resizable hash table API for libnl.
+ */
+
+#ifndef NETLINK_HASHTABLE_API_H_
```
thom311, line 6: as `include/nl-priv-dynamic-core/README.md` tries to explain, this directory contains headers that are internal API, however, they are exposed by `libnl-3.so`.

For example, it provides `__list_str2type()`. This is not public API, but to be usable to other libraries like `libnl-route-3.so`. See `git grep __list_str2type libnl-3.sym` and `nm ./lib/.libs/libnl-3.so | grep __list_str2type`.

Point being, `hashtable-api.h` does not fit there.

But it is also only used by `lib/{cache,hashtable}.c`.

The simplest solution is to move `include/nl-priv-dynamic-core/hashtable-api.h` to `lib/hashtable-api.h`.

cpaasch-oai: Thanks for this feedback! It was indeed not very clear to me how this actually plays out.

I'm changing this.

PR #435, opened 2025-07-28
[nexthop] Add a bunch of setters/getters to for fields in struct rtnl_nh

cpaasch-oai:
This pull-request further works towards an implementation where libnl can push nexthop configuration similar to `ip nexthop` down to the kernel.

We need a bunch of setters to complete this.

The `rtnl_nh_add` to actually create the netlink message comes in a follow-up PR.


lib/route/nh.c
```diff
@@ -221,6 +221,20 @@ int rtnl_nh_get_oif(struct rtnl_nh *nexthop)
 	return 0;
 }
 
+int rtnl_nh_set_oif(struct rtnl_nh *nexthop, int ifindex)
+{
+	if (!nexthop)
+		return -NLE_INVAL;
+
+	if (ifindex <= 0)
+		return -NLE_RANGE;
```
thom311, line 246: the default for the value is zero. So the setter should allow to reset the default.

It's only unclear, whether setting to 0 should also mean to unset NH_ATTR_OIF in the mask.

I think kernel would not differentiate between unset and (explictly set to) zero. Except for older kernels that didn't support NHA_OIF yet. But I think we should not take that as indication that libnl should differenciate. Meaning: explicitly setting to zero should be allowed, but it also means to clear NH_ATTR_OIF.

WDYT?

lib/route/nh.c
```diff
@@ -273,6 +287,20 @@ int rtnl_nh_get_id(struct rtnl_nh *nh)
 	return -NLE_INVAL;
 }
 
+int rtnl_nh_set_id(struct rtnl_nh *nh, uint32_t id)
+{
+	if (!nh)
+		return -NLE_INVAL;
+
+	if (id == 0)
+		return -NLE_INVAL;
```
thom311, line 466: same here. Setting to zero should be allowed, an unset `ce_mask`.

lib/route/nh.c
```diff
@@ -240,6 +240,24 @@ int rtnl_nh_get_fdb(struct rtnl_nh *nexthop)
...
 }
 
+int rtnl_nh_set_family(struct rtnl_nh *nexthop, uint8_t family)
+{
+	if (!nexthop)
+		return -NLE_INVAL;
+
+	nexthop->nh_family = family;
+
+	return 0;
```
thom311, line 266: seems `nh_compare()` ignores `nh_family`. That should be fixed... (not necessarily by you).

libnl-route-3.sym
```diff
@@ -1372,6 +1372,7 @@ global:
 	rtnl_link_is_bond;
 	rtnl_nh_get_family;
 	rtnl_nh_get_oif;
+	rtnl_nh_set_group;
```
thom311, line 1379: please sort.

lib/route/nh.c
```diff
@@ -258,6 +258,34 @@ int rtnl_nh_get_family(struct rtnl_nh *nexthop)
 	return nexthop->nh_family;
 }
 
+int rtnl_nh_set_group(struct rtnl_nh *nexthop, nl_nh_group_info_t *entries,
```
thom311, line 385: I think this should be a `const nl_nh_group_info_t *` pointer.

lib/route/nh.c
```diff
@@ -258,6 +258,34 @@ int rtnl_nh_get_family(struct rtnl_nh *nexthop)
...
+	if (size == 0)
+		return -NLE_INVAL;
+
+	nhg = rtnl_nh_grp_alloc(size);
+	if (!nhg)
+		return -NLE_NOMEM;
+
+	for (unsigned i = 0; i < size; i++) {
+		nhg->entries[i].nh_id = entries[i].nh_id;
+		nhg->entries[i].weight = entries[i].weight;
```
thom311, line 402: maybe `memcpy()`?

lib/route/nh.c
```diff
@@ -258,6 +258,34 @@ int rtnl_nh_get_family(struct rtnl_nh *nexthop)
...
+int rtnl_nh_set_group(struct rtnl_nh *nexthop, nl_nh_group_info_t *entries,
+		      unsigned size)
+{
+	nl_nh_group_t *nhg = NULL;
+
+	if (!nexthop || !entries)
+		return -NLE_INVAL;
+
+	if (size == 0)
+		return -NLE_INVAL;
```
thom311, line 394: is there a problem with allowing to set this to zero?

lib/route/nh.c
```diff
@@ -135,6 +136,7 @@ static int nh_clone(struct nl_object *_src, struct nl_object *_dst)
 	dst->nh_family = src->nh_family;
 	dst->nh_id = src->nh_id;
 	dst->nh_oif = src->nh_oif;
+	dst->nh_group_type = src->nh_group_type;
```
thom311, line 150: also missing in `nh_compare()`?

lib/route/nh.c
```diff
@@ -258,6 +260,28 @@ int rtnl_nh_get_family(struct rtnl_nh *nexthop)
...
+
+int rtnl_nh_get_group_type(struct rtnl_nh *nexthop)
+{
+	if (!nexthop)
+		return -NLE_INVAL;
+
+	if (!(nexthop->ce_mask & NH_ATTR_GROUP_TYPE))
+		return -NLE_INVAL;
+
+	return (int)nexthop->nh_group_type;
```
thom311, line 296: in this case, we encode the error value and the success value in `int`. I think that is fine for `NHA_GROUP_TYPE`.

At least, as long as we are sure that this value is always just an `u16` (that fits inside the positive range of `int`).

In that case, `nh_group_type` in `struct rtnl_nh` is wrongly defined as `uint32_t`. Could you please fix that too?

Or do you think there might be cases where it's not `u16`? In that case, the getter would have to return the value and the error code separately.

include/netlink/route/nh.h
```diff
@@ -56,6 +56,19 @@ extern int rtnl_nh_get_family(struct rtnl_nh *);
...
+					     uint16_t buckets);
+extern int rtnl_nh_get_res_group_bucket_size(struct rtnl_nh *);
+
+extern int rtnl_nh_set_res_group_idle_timer(struct rtnl_nh *,
+					    uint32_t idle_timer);
+extern int rtnl_nh_get_res_group_idle_timer(struct rtnl_nh *);
+
+extern int rtnl_nh_set_res_group_unbalanced_timer(struct rtnl_nh *,
+						  uint32_t unbalanced_timer);
+extern int rtnl_nh_get_res_group_unbalanced_timer(struct rtnl_nh *);
```
thom311, line 70: for `rtnl_nh_get_res_group_idle_timer()` and `rtnl_nh_get_res_group_unbalanced_timer()` it seems that the full range of values is `uint32_t`.

We cannot encode that in an `int` return value. The getters need to have a separate `uint32_t *out_value` parameter.

lib/route/nh.c
```diff
@@ -452,6 +552,33 @@ static int nexthop_msg_parser(struct nl_cache_ops *ops, struct sockaddr_nl *who,
...
+				nexthop->ce_mask |= NH_ATTR_RES_BUCKETS;
+			}
+			if (rg[NHA_RES_GROUP_IDLE_TIMER]) {
+				nexthop->res_grp_idle_timer = nla_get_u32(
+					rg[NHA_RES_GROUP_IDLE_TIMER]);
+				nexthop->ce_mask |= NH_ATTR_RES_IDLE_TIMER;
+			}
+			if (rg[NHA_RES_GROUP_UNBALANCED_TIMER]) {
+				nexthop->res_grp_unbalanced_timer = nla_get_u32(
+					rg[NHA_RES_GROUP_UNBALANCED_TIMER]);
```
thom311, line 573: you should pass a `policy` argument to `nla_parse_nested()`.

Otherwise, you cannot blindly trust that the provided value is in fact a `nla_get_u32()`.

Theoretically, you could also check each attribute individually. But seems simpler to create a policy.

cpaasch-oai:
Actually - as I am building a lot of tests for rtnl_nh I realize that things are inconsistent across the board regarding input-checks (beyond the code I'm adding here). Sometimes there is :
```
int rtnl_rule_get_l3mdev(struct rtnl_rule *rule)
{
        if (!rule)
                return -NLE_INVAL;
```

And in other cases there is no null-check at all.

Do you think we should unify this ? IMO I would say that no null-check is better to avoid silent failures. But well - it's a judgement call :)

thom311:
merged, thank you!

PR #437, opened 2025-07-30
Add extensive tests for ip6tnl

cpaasch-oai:
This PR adds extensive tests to all setters/getters of the ip6 link tunnel configuration.

While we are at it, opt-int ip6tnl files to clang-format.sh and fix a typo in nl-test-util.

### DIFF
--- a/tests/nl-test-util.c
+++ b/tests/nl-test-util.c
@@ -731,7 +731,7 @@ void _nltst_assert_link_exists_full(const char *ifname, bool exists)
 
 			ck_assert_msg(
 				0,
-				"link(%s) does not exist (stat(%s) failed with r=%d, errno=%d (%s)",
+				"link(%s) does not exist (stat(%s) failed with r=%d, errno=%d (%s))",
 				ifname, path, r, errsv, strerror(errsv));
 		}
 	} else {


tests/cksuite-link-ip6tnl.c
```diff
@@ -0,0 +1,270 @@
...
+#include <string.h>
+#include <netlink/errno.h>
+#include <netlink/route/link/ip6tnl.h>
+
+static void _delete_link(const char *ifname)
+{
+	_nltst_delete_link(NULL, ifname);
+}
+#define _nltst_auto_delete_link _nl_auto(_nltst_auto_delete_link_fcn)
+_NL_AUTO_DEFINE_FCN_TYPED0(const char *, _nltst_auto_delete_link_fcn,
```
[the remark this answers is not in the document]
thom311: There is also a problem here. `_nltst_delete_link()` asserts that `ifname` is not NULL and that such a link exists.

At the very least, the `_nl_auto()` macro should be graceful against `ifname` NULL.

But I wonder, maybe it should also be graceful against missing interfaces. That is unclear. It probably is preferable to be strict here...

tests/cksuite-link-ip6tnl.c
```diff
@@ -0,0 +1,270 @@
...
+#include "cksuite-all.h"
+#include "nl-aux-route/nl-route.h"
+#include "nl-default.h"
+#include "nl-test-util.h"
+
+#include <arpa/inet.h>
+#include <check.h>
+#include <string.h>
+#include <netlink/errno.h>
+#include <netlink/route/link/ip6tnl.h>
```
thom311, line 22: includes should always start with

```
#include "nl-default.h"
```

This pulls in some default things that should be available everywhere, including `config.h` as first (which is kinda important to always do).

Then we do standard headers `<>`.

Then, we do other libnl headers with `<>`.

Then, internal libnl heades with `""`.

Finally, we do headers from `tests`, like `cksuite-all.h`.

(if that order is violated somewhere, it is a minor bug).

tests/cksuite-link-ip6tnl.c
```diff
@@ -0,0 +1,270 @@
...
+	auto_del_dummy = IFNAME_DUMMY;
+
+	rtnl_link_set_name(link, IFNAME);
+
+	_nltst_add_link(sk, IFNAME_DUMMY, "dummy", &ifindex_dummy);
+	r = rtnl_link_ip6_tnl_set_link(link, ifindex_dummy);
+	ck_assert_int_eq(r, 0);
+
+	inet_pton(AF_INET6, "2001:db8::10", &addr_local);
+	inet_pton(AF_INET6, "2001:db8::20", &addr_remote);
```
thom311, line 79: `addr_local = _nltst_inet6("2001:db8::10")`

thom311 (changes requested):

.gitignore
```diff
@@ -128,7 +128,6 @@ test-suite.log
 /tests/test-create-bridge
 /tests/test-create-geneve
 /tests/test-create-ifb
-/tests/test-create-ip6tnl
```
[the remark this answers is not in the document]
cpaasch-oai: Will do!

tests/cksuite-link-ip6tnl.c
cpaasch-oai: Thanks for clarifying that! I fixed this.

cpaasch-oai:
> thank you for adding tests. That is very good!

Cool! I will add more. It goes quick.

PR #438, opened 2025-07-31
lib: avoid overflow in computation of s_seq_expect

cferris1000:
On some systems, the clock is reset, or is lost, so the value returned by the time function can be a very small value. In that case, the _badrandom_from_time functions returns a large value close to the maximum unsigned int value for s_seq_expect. This can lead to the value wrapping around fairly quickly.

When compiling the library with the unsigned-integer-overflow sanitizer enabled, this causes an abort.

Detect this potential wrap around condition and avoid it.

### DIFF
--- a/lib/nl.c
+++ b/lib/nl.c
@@ -896,7 +896,10 @@ continue_reading:
 		    hdr->nlmsg_type == NLMSG_OVERRUN) {
 			/* We can't check for !NLM_F_MULTI since some netlink
 			 * users in the kernel are broken. */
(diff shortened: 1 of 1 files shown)


thom311:
Unsigned overflow is explicitly supported and fine in C. And obviously, the overflow here is not an accident or unintentional.

It seems to me that this sanitizer is a bit too strict. Can you even enable it on a non-trivial code base? Isn't unsigned overflow a common thing (that often enough happens intentionally)?

Anyway. Will merge. Thank you.

PR #440, opened 2025-08-01
Minor fixes to nh.c

cpaasch-oai:
While working on the PR-feedback from https://github.com/thom311/libnl/pull/435 I found a set of bugs unrelated to the work in there. So, sending out a separate PR for these fixes.

(I should spend some time to create a full set of tests for the API from nh.h)


thom311:
merged as 92cc31a6ad61fe61834042d6172c632aa53e6d81

thank you!

PR #441, opened 2025-08-04
Add rtnl_nh_add() API

cpaasch-oai:
The goal of this PR is to support pushing a rtnl_nh down to the kernel through netlink, similar to what `ip nexthop add` does.

A few additional things in this PR (1 bugfix, some more tests for getters/setters and adding support for ENETDOWN). Let me know if I should split it in separate PRs - I don't want to create overhead for you @thom311 .

f4d68e962547e1896c73d005ffcc068bb9abd476 adds an extensive testsuite for the rtnl_nh getters/setters. All in user-space. As part of adding this work, one bug was identified (fixed with 87c7af13483ce1f86357c1db9724c7ccbe294bbf).

79d1abd230132b15e0cd2bf869cf8036b6882617 adds another error-code for ENETDOWN, which turned out to be useful when testing the kernel round-trip for rtnl_nh_add.

933edc31daec4d00bdf38d77ee54bef251622f23 actually adds the rtnl_nh_add() call.

c46fa32713abe066075fccddfe481225c9b30809 tries to be a complete test-suite for the rtnl_nh_add() call as well as fetching nexthop objects (and parsing the netlink message) from the kernel.

### DIFF
--- a/include/nl-aux-route/nl-route.h
+++ b/include/nl-aux-route/nl-route.h
@@ -13,6 +13,12 @@ void rtnl_link_put(struct rtnl_link *);
 _NL_AUTO_DEFINE_FCN_TYPED0(struct rtnl_link *, _nl_auto_rtnl_link_fcn,
 			   rtnl_link_put);
 
+struct rtnl_addr;
+void rtnl_addr_put(struct rtnl_addr *);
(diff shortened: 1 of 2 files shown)


thom311:
thank you for all the tests. This is really nice!!

PR #442, opened 2025-08-06
Set of minor fixes across the board

cpaasch-oai:
While working on additional tests for nexthop encapsulation, I realized there are some bugs in certain places across libnl.

I'm addressing quite a few of these with this PR.

Tests for each of these would be great, but it takes a bit more time. Thus, I fix it first and will add tests if I have time.

### DIFF
--- a/lib/route/link/ip6gre.c
+++ b/lib/route/link/ip6gre.c
@@ -26,6 +26,7 @@
 
 #include "nl-route.h"
 #include "link-api.h"
+#include "nl-aux-route/nl-route.h"
 
 #define IP6GRE_ATTR_LINK          (1 << 0)
 #define IP6GRE_ATTR_IFLAGS        (1 << 1)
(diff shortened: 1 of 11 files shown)


lib/route/link/ip6gre.c
```diff
@@ -246,8 +246,15 @@ static void ip6gre_dump_details(struct rtnl_link *link, struct nl_dump_params *p
 	char addr[INET6_ADDRSTRLEN];
 
 	if (ip6gre->ip6gre_mask & IP6GRE_ATTR_LINK) {
+		struct rtnl_link *parent;
+
 		nl_dump(p, "      link ");
-		name = rtnl_link_get_name(link);
+
+		name = NULL;
+		parent = link_lookup(link->ce_cache, ip6gre->link);
```
thom311, line 255: I also think that `link_lookup()` returns a reference. So you would need `_nl_auto_rtnl_link`.

On the other hand, `link_lookup()` is internal API only (it can change). This API shouldn't return a reference/ownership, that would only make sense if cache and `nl_object_get()` were thread-safe and could be filled at the same time by other threads. But it is not, so nobody is going to unref the returned object unexpectedly. The callers just need to take care that they do nothing with the looked up link, after it might be destroyed (or take a reference here).

I would even go a step further and add a `link_lookup_name()` helper (that returns a `const char *`) and use that.

cpaasch-oai: Yes, we are leaking objects. [75352b7](https://github.com/thom311/libnl/pull/442/commits/75352b796993b264879dd0469325925b09e4c143) intended to fix that.

If you rather prefer to change `link_lookup` to not take the ref and/or add `link_lookup_name`, I can do that !

Issue #444, opened 2025-08-07
test_kernel_roundtrip_group_resilient()'s success depends on kernel HZ


KanjiMonster:
`test_kernel_roundtrip_group_resilient()` sets timers and compares the set values after retrieving the objects from the kernel.

Unfortunately the timer values returned may differ depending on the configured `HZ` value of the running kernel. The kernel does not store the raw timer values in its database, but converts them to jiffies first, which may not convert cleanly.

I noticed this on Manjaro/arch, where the kernel uses `CONFIG_HZ=300`, which causes the conversion from seconds to jiffies to seconds to have a loss of precision, while e.g. Ubuntu with `CONFIG_HZ=1000` is fine.

Grepping through the kernel tree, depending on the arch there are quite a few options for HZ, so I tested the values wether the conversion and back is neutral, and for most it isn't:

(seconds/input -> seconds to jiffies -> jiffies to seconds/output, with a ! if it doesn't match the initial value)

```
CONFIG_HZ=  24 => 15 ->   3 -> 12! 25 ->   6 -> 25 
CONFIG_HZ=  32 => 15 ->   4 -> 12! 25 ->   8 -> 25 
CONFIG_HZ=  48 => 15 ->   7 -> 14! 25 ->  12 -> 24!
CONFIG_HZ= 100 => 15 ->  15 -> 15  25 ->  25 -> 25 
CONFIG_HZ= 128 => 15 ->  19 -> 14! 25 ->  32 -> 25 
CONFIG_HZ= 250 => 15 ->  37 -> 14! 25 ->  62 -> 24!
CONFIG_HZ= 256 => 15 ->  38 -> 14! 25 ->  64 -> 25 
CONFIG_HZ= 300 => 15 ->  45 -> 14! 25 ->  75 -> 24!
CONFIG_HZ= 500 => 15 ->  75 -> 15  25 -> 125 -> 25 
CONFIG_HZ=1000 => 15 -> 150 -> 15  25 -> 250 -> 25 
CONFIG_HZ=1024 => 15 -> 153 -> 14! 25 -> 256 -> 25 
CONFIG_HZ=1200 => 15 -> 180 -> 14! 25 -> 300 -> 24!
```

So the only values working fine are 100, 500, and 1000.

Not sure if this counts as a kernel bug or if this is expected/allowed.

thom311:
fixed by 69231b6e8ef1c80ac51d7327701a827d27ba4dac .

PR #445, opened 2025-08-08
[th/tests-relax-hz-timers] route/tests: relax check for nh timers

thom311:
Kernel stores the timer values internally in HZ size, so it looses precision when converting to HZ and back. And as the HZ value is configurable, it depends on the kernel, how exactly the values get mangled.

Relax the check for that.

https://github.com/thom311/libnl/issues/444


thom311:
merged as 69231b6e8ef1c80ac51d7327701a827d27ba4dac

